PR #2, opened 2019-07-08
Explicit Strong Form Class Added (not complete)

abtin98:


dougshidong:
No need to derive DGStrongExplicit.

PR #83, opened 2021-09-27
Navier-Stokes higher-order convergence

jbrillon:
This pull request is to incorporate the minor additions to convergence studies made while I investigated the higher-order convergence of Navier-Stokes by running cases on the Beluga cluster.  

These minor changes include:
- Additional parameter,  `output_solution`, introduced for `manufactured_convergnce_study_parameters`, which gives the option to output the solution for any test using `grid_study.cpp` 
- Added the `diss_num_flux_string` to the outputted convergence tables filename
- Convergence tables are now outputted as we loop through the grids in `grid_study.cpp`


src/testing/grid_study.cpp
```diff
@@ -186,6 +188,62 @@ ::run_test () const
     std::vector<double> fail_conv_slop;
     std::vector<dealii::ConvergenceTable> convergence_table_vector;
 
+    // Write the convergence table to a file // add these to the parameter files as member functions /// look into creating a dictionary/map for this so were not double hard coding it
```
dougshidong, line 191: Can you move this to a separate function?

jbrillon: Yes I will do that, good call

PR #85, opened 2021-10-07
Reduced order methods

donovan97:
Pull request to add reduced-order methods POD-Galerkin and POD-Petrov Galerkin, with test on Burger's Rewienski. 

The changes are the following:
- Addition of `PODGalerkinODESolver` and `PODPetrovGalerkinODESolver` classes, derived from `ODESolverBase`. 
- Addition of 3 new factory methods in ODESolverFactory. One factory takes only a `DGBase` and can create an `ImplicitODESolver` or an `ExplicitODESolver`. One factory takes a `DGBase` and `POD` and can create a `PODGalerkinODESolver` and `PODPetrovGalerkinODESolver`. The other two factories are identical except they also take a `ODESolverEnum` allowing the user to specify exactly which solver is requested (instead of choosing value specified in the parameter).
- pod_basis with `POD` is changed to a `TrilinosWrapper::SparseMatrix` to interface with the ODE solver classes.
- Creation of `ReducedOrder` test which first runs 3 snapshots. This is done within the Cmake file, by calling the `BurgersRewienskiSnapshot` test. The `ReducedOrder` test uses the snapshots to generate a POD basis. All basis functions are used to ensure that the reduced order solution matches the full order solution exactly. If solutions do not match, test fails.


dougshidong (approved):
Looking good. Would it be possible to provide references to some of the papers you might have used for the POD basis? For example, whenever some code closely related to a formula of a paper, mention paper and formula number.

donovan97:
> Looking good. Would it be possible to provide references to some of the papers you might have used for the POD basis? For example, whenever some code closely related to a formula of a paper, mention paper and formula number.

Done in the most recent commit.

PR #88, opened 2021-11-12
Reduced order

donovan97:
Pull request to incorporate the addition of adjoint-based POD adaptation.

The changes include:

- Replacement of `src/pod` directory with `src/reduced_order` directory
- Renaming of `proper_orthogonal_decomposition.cpp` to `pod_basis.cpp`
- Addition of functions within class `POD` to: write POD basis to a file and read later on, read POD basis from file, build `TrilinosWrappers::SparseMatrix` of the full POD basis. Also added capacity to compute POD basis using the method of snapshots, which requires taking the SVD on a smaller matrix than the simple method of taking it directly on the matrix of snapshots. 
- Addition of derived class `SpecificPOD` that groups common elements of further derived classes `CoarsePOD`, `FinePOD`, `FineNotInCoarsePOD`. `SpecificPOD` contains function to add basis columns to construct and modify the POD bases. `FineNotInCoarsePOD` contains function to remove basis functions as the coarse basis is refined.
- Addition of `PODAdaptation` class which implements simple and progressive POD adaptation. This class computes the fine-space adjoint, the coarse-space residual and the dual-weighted residual. Also contains the function `getPODBasisColumnsToAdd` which decides which POD basis vectors to add to the coarse basis, based on user-specified options (fixed number of basis vectors or automatic choosing if considering the sign of the error.)
- Removal of test `BurgersRewienskiSnapshot`. This is now integrated with the `FlowSolver`. `BurgersRewienskiSnapshot`  was added as a flow solver case in `/src/testing/flow_solver_cases/`
- Modifications to `FlowSolver` to add ability to compute steady-state solutions.
- Addition of various parameters to `parameters_reduced_order.cpp`
- Modifications to `PODGalerkinODESolver` and `PODPetrovGalerkinODESolver` to use pointers to matrices
- Reorganization and renaming of tests in `/tests/reduced_order/`
- Addition of new test `1D_BURGERS_REWIENSKI_POD_ADAPTATION` which performs a POD adaptation and then verifies that the coarse basis meets the error tolerance specified in the parameter file.
- Simplification of existing `1D_BURGERS_REWIENSKI_REDUCED_ORDER` tests to greatly reduce test time.


dougshidong:
Went through like half. Will continue tomorrow. It's a huge PR.

donovan97:
> Went through like half. Will continue tomorrow. It's a huge PR. I suggest breaking it down into smaller tasks for future PR.

Yeah sorry about that, it got a bit out of hand. I will try to make smaller PR going forward.

src/testing/reduced_order_pod_adaptation.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+template <int dim, int nstate, typename real>
+template <typename real2>
+real2 BurgersRewienskiFunctional<dim,nstate,real>::evaluate_volume_integrand(
+    const PHiLiP::Physics::PhysicsBase<dim,nstate,real2> &/*physics*/,
+    const dealii::Point<dim,real2> &/*phys_coord*/,
+    const std::array<real2,nstate> &soln_at_q,
+    const std::array<dealii::Tensor<1,dim,real2>,nstate> &/*soln_grad_at_q*/) const
+{
```
jbrillon, line 111: Indent contents of this function

src/testing/reduced_order_pod_adaptation.h
```diff
@@ -0,0 +1,70 @@
...
+#include "tests.h"
+#include "dg/dg.h"
+#include "parameters/all_parameters.h"
+#include "functional/functional.h"
+
+namespace PHiLiP {
+namespace Tests {
+
+/// Burgers Rewienski snapshot
+    template <int dim, int nstate>
```
jbrillon, line 13: Untab everything once

src/reduced_order/pod_basis.cpp
```diff
@@ -0,0 +1,261 @@
+#include <deal.II/lac/sparse_matrix.h>
+#include "pod_basis.h"
+#include <deal.II/fe/mapping_q1_eulerian.h>
+
+namespace PHiLiP {
+namespace ProperOrthogonalDecomposition {
+
+template <int dim>
+POD<dim>::POD(std::shared_ptr<DGBase<dim,double>> &_dg)
```
jbrillon, line 9: Same as the previous comment on the underscore notation for inputs

src/reduced_order/pod_basis.cpp
```diff
@@ -0,0 +1,261 @@
...
+            int cols = 0;
+            //First loop set to count rows and columns
+            while(std::getline(myfile, line)){ //for each line
+            std::istringstream stream(line);
+            std::string field;
+            cols = 0;
+            while (getline(stream, field,' ')){ //parse data values on each line
+                if (field.empty()){ //due to whitespace
+                    continue;
+                }else{
```
jbrillon, line 188:
```suggestion
                } else {
```

src/reduced_order/pod_basis.cpp
```diff
@@ -0,0 +1,261 @@
...
+            int cols = 0;
+            //First loop set to count rows and columns
+            while(std::getline(myfile, line)){ //for each line
+                std::istringstream stream(line);
+                std::string field;
+                cols = 0;
+                while (getline(stream, field,' ')){ //parse data values on each line
+                    if (field.empty()){ //due to whitespace
+                        continue;
+                    }else{
```
jbrillon, line 53:
```suggestion
                    } else {
```

src/reduced_order/pod_basis_types.cpp
```diff
@@ -0,0 +1,124 @@
+#include "pod_basis_types.h"
+
+namespace PHiLiP {
+namespace ProperOrthogonalDecomposition {
+
+template <int dim>
+SpecificPOD<dim>::SpecificPOD(std::shared_ptr<DGBase<dim,double>> &_dg)
```
jbrillon, line 7: Same comment regarding the underscore notation

jbrillon (commented):
Great work! Mostly left some minor comments, other than the introduction of that intermediate class I mention that could be a good idea.

src/reduced_order/pod_adaptation.cpp
```diff
@@ -0,0 +1,197 @@
...
+    dealii::ParameterHandler parameter_handler;
+    Parameters::LinearSolverParam::declare_parameters (parameter_handler);
+    this->linear_solver_param.parse_parameters (parameter_handler);
+}
+
+template <int dim, int nstate>
+void PODAdaptation<dim, nstate>::progressivePODAdaptation()
+{
+    getDualWeightedResidual();
+    while(abs(error) > all_parameters->reduced_order_param.adaptation_tolerance){
```
dougshidong, line 28: If that guaranteed to terminate?

src/ode_solver/pod_petrov_galerkin_ode_solver.h
```diff
@@ -35,16 +35,16 @@ class PODPetrovGalerkinODESolver: public ODESolverBase <dim, real, MeshType>
...
-    dealii::LinearAlgebra::distributed::Vector<double> reduced_rhs;
+    std::shared_ptr<dealii::LinearAlgebra::distributed::Vector<double>> reduced_rhs;
 
     /// Psi = J * V
-    dealii::TrilinosWrappers::SparseMatrix psi;
+    std::shared_ptr<dealii::TrilinosWrappers::SparseMatrix> psi;
 
     /// Reduced lhs for linear solver
-    dealii::TrilinosWrappers::SparseMatrix reduced_lhs;
+    std::shared_ptr<dealii::TrilinosWrappers::SparseMatrix> reduced_lhs;
```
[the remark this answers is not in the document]
donovan97: It's not a bad idea, but the only thing that is really common to both classes is the `allocate_ode_system` functions. But even then, these are not exactly the same and would need to override an intermediate class. The intermediate class would probably also need to be in its own file for clarity, so I don't know that it's really worth making a new class for fairly little gain.

src/ode_solver/pod_galerkin_ode_solver.h
```diff
@@ -16,14 +16,14 @@ template <int dim, typename real, typename MeshType = dealii::Triangulation<dim>
 #else
 template <int dim, typename real, typename MeshType = dealii::parallel::distributed::Triangulation<dim>>
 #endif
-class PODGalerkinODESolver: public ODESolverBase <dim, real, MeshType>
+class PODGalerkinODESolver: public ImplicitODESolver<dim, real, MeshType>
```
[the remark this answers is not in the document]
donovan97: This is a pretty tricky one. Looking at it from the point of view of has-a vs is-a, I would argue that the Galerkin and Petrov-Galerkin ODE solvers are an implicit ODE solver. They are fundamentally using the same implicit scheme to solve the equation, but with a matrix multiplication to project to a different subspace. Once they are created using a POD basis, they can be used interchangeably with an implicit solver anywhere in the code without any issue which is the main idea behind the LSP. The only thing is that the solution obtained would be different, or if a bad POD basis is used, the solution might not converge. Therefore, in that sense, the preconditions are different and the reduced-order implicit solvers require a POD basis that properly describes the solution space. So the LSP might not be strictly respected. 

However, the constructors for the reduced-order implicit solvers require a POD basis. Therefore, it would have to be assumed that if a reduced-order implicit solver is being created and used in the place of an implicit solver, the user is fully aware of the preconditions. Otherwise, the reduced-order implicit solver wouldn't be getting created in the first place. The same goes for the factory class, it's not possible to create a reduced-order solver unless you're giving it a POD basis.

Owning an ImplicitODESolver brings a series of new problems because of the current inheritance-based design. In the reduced-order solver wouldn't be able to interface with the private or protected members of the implicit solver in its current form. It would have been possible to make a composition-based solution, but that would likely involve re-writing the entire ODE solver section.

jbrillon (approved):

src/ode_solver/pod_petrov_galerkin_ode_solver.h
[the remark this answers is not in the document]
jbrillon: @dougshidong What's your take on this? Is an intermediate class for just 5 member variables overkill? I'm wondering just in general, whats the minimum merit for an intermediate class

src/ode_solver/ode_solver_base.cpp
```diff
@@ -176,6 +176,17 @@ int ODESolverBase<dim,real,MeshType>::steady_state ()
         if(CFL_factor <= 1e-2) this->dg->right_hand_side.add(1.0);
     }
 
+    if (ode_param.output_solution_vector_modulo > 0) {
+        for (unsigned int i = 0; i < this->dg->solution.size(); ++i) {
+            solutions_table.add_value(
+                    "Steady-state solution:",
+                    this->dg->solution[i]);
+        }
+        solutions_table.set_precision("Steady-state solution:", 8);
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit

src/reduced_order/pod_adaptation.cpp
```diff
@@ -0,0 +1,197 @@
+#include "pod_adaptation.h"
+
+namespace PHiLiP {
+namespace ProperOrthogonalDecomposition {
+
+using DealiiVector = dealii::LinearAlgebra::distributed::Vector<double>;
+
+template <int dim, int nstate>
+PODAdaptation<dim, nstate>::PODAdaptation(std::shared_ptr<DGBase<dim,double>> &_dg, Functional<dim,nstate,double> &_functional)
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/testing/flow_solver.cpp
```diff
@@ -125,36 +126,55 @@ int FlowSolver<dim,nstate>::run_test() const
...
+        // Steady-state solution
+        //----------------------------------------------------
+        ode_solver->steady_state();
+    }
     pcout << "done." << std::endl;
     return 0;
 }
+
+template <int dim, int nstate>
+double FlowSolver<dim,nstate>::get_constant_time_step(std::shared_ptr<DGBase<dim,double>> /*dg*/) const
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/testing/reduced_order_pod_adaptation.h
donovan97: Fixed in most recent commit.

src/reduced_order/pod_basis.cpp
donovan97: Fixed in most recent commit.

src/reduced_order/pod_basis_types.cpp
donovan97: Fixed in most recent commit.

src/ode_solver/pod_petrov_galerkin_ode_solver.h
```diff
@@ -16,14 +16,14 @@ template <int dim, typename real, typename MeshType = dealii::Triangulation<dim>
...
+class PODPetrovGalerkinODESolver: public ImplicitODESolver<dim, real, MeshType>
 {
 public:
     /// Default constructor that will set the constants.
-    PODPetrovGalerkinODESolver(std::shared_ptr< DGBase<dim, real, MeshType> > dg_input, std::shared_ptr<ProperOrthogonalDecomposition::POD> pod); ///< Constructor.
+    PODPetrovGalerkinODESolver(std::shared_ptr< DGBase<dim, real, MeshType> > dg_input, std::shared_ptr<ProperOrthogonalDecomposition::POD<dim>> pod); ///< Constructor.
 
     ///POD
-    std::shared_ptr<ProperOrthogonalDecomposition::POD> pod;
+    std::shared_ptr<ProperOrthogonalDecomposition::POD<dim>> pod;
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit

src/reduced_order/pod_adaptation.h
```diff
@@ -0,0 +1,100 @@
...
+    Functional<dim,nstate,double> &functional;
+
+    /// Smart pointer to DGBase
+    std::shared_ptr<DGBase<dim,double>> dg;
+
+    /// Pointer to all parameters
+    const Parameters::AllParameters *const all_parameters;
+
+    /// Smart pointer to coarse POD basis
+    std::shared_ptr<ProperOrthogonalDecomposition::CoarsePOD<dim>> coarsePOD;
```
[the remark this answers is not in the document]
donovan97: In this case, the CoarsePOD cannot be made unique since it is passed to the ODE solver. However, I made the other POD bases unique_ptr.

src/testing/reduced_order_pod_adaptation.cpp
```diff
@@ -0,0 +1,128 @@
...
+real2 val = 0;
+
+// integrating over the domain
+for (int istate=0; istate<nstate; ++istate) {
+    val += soln_at_q[istate];
+}
+
+return val;
+}
```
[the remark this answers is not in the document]
donovan97: Comment added in most recent commit.

src/reduced_order/pod_adaptation.cpp
donovan97: It was not, I have fixed this in the most recent commit.

src/ode_solver/pod_petrov_galerkin_ode_solver.h
```diff
@@ -35,16 +35,16 @@ class PODPetrovGalerkinODESolver: public ODESolverBase <dim, real, MeshType>
...
-    dealii::LinearAlgebra::distributed::Vector<double> reduced_solution_update;
+    std::shared_ptr<dealii::LinearAlgebra::distributed::Vector<double>> reduced_solution_update;
 
     /// Reduced rhs for linear solver
-    dealii::LinearAlgebra::distributed::Vector<double> reduced_rhs;
+    std::shared_ptr<dealii::LinearAlgebra::distributed::Vector<double>> reduced_rhs;
 
     /// Psi = J * V
-    dealii::TrilinosWrappers::SparseMatrix psi;
+    std::shared_ptr<dealii::TrilinosWrappers::SparseMatrix> psi;
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/testing/flow_solver_cases/1D_burgers_rewienski_snapshot.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+}
+
+template <int dim, int nstate>
+void BurgersRewienskiSnapshot<dim,nstate>::display_flow_solver_setup() const
+{
+    using PDE_enum = Parameters::AllParameters::PartialDifferentialEquation;
+    const PDE_enum pde_type = this->all_param.pde_type;
+    std::string pde_string;
+    if (pde_type == PDE_enum::burgers_rewienski)    {pde_string = "burgers_rewienski";}
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/testing/reduced_order_pod_adaptation.h
```diff
@@ -10,60 +10,60 @@ namespace PHiLiP {
...
+    using FadFadType = Sacado::Fad::DFad<FadType>; ///< Sacado AD type that allows 2nd derivatives.
+public:
+    /// Constructor
+    BurgersRewienskiFunctional(
+            std::shared_ptr<PHiLiP::DGBase<dim,real>> dg_input,
+            std::shared_ptr<PHiLiP::Physics::PhysicsBase<dim,nstate,FadFadType>> _physics_fad_fad,
+            const bool uses_solution_values = true,
+            const bool uses_solution_gradient = false)
+            : PHiLiP::Functional<dim,nstate,real>(dg_input,_physics_fad_fad,uses_solution_values,uses_solution_gradient)
+    {}
```
jbrillon, line 39: Shouldn't there be a semi-colon here after `{}` ?

src/testing/reduced_order_pod_adaptation.h
```diff
@@ -10,60 +10,60 @@ namespace PHiLiP {
...
+        return evaluate_volume_integrand<>(physics, phys_coord, soln_at_q, soln_grad_at_q);
+    }
+    /// Non-template functions to override the template classes
+    FadFadType evaluate_volume_integrand(
+            const PHiLiP::Physics::PhysicsBase<dim,nstate,FadFadType> &physics,
+            const dealii::Point<dim,FadFadType> &phys_coord,
+            const std::array<FadFadType,nstate> &soln_at_q,
+            const std::array<dealii::Tensor<1,dim,FadFadType>,nstate> &soln_grad_at_q) const override
+    {
+        return evaluate_volume_integrand<>(physics, phys_coord, soln_at_q, soln_grad_at_q);
```
jbrillon, line 64: why the empty `<>` ?

tests/integration_tests_control_files/reduced_order/pod_adaptation_snapshot_generator.sh
```diff
@@ -63,7 +63,7 @@ echo "end"
 
 
 dir=$(pwd)
-/usr/bin/mpirun "-n" "1" "$HOME/Codes/PHiLiP/cmake-build-release/bin/PHiLiP_1D" "-i" "${dir}/${file}"
+/usr/bin/mpirun "-n" "1" "$1/PHiLiP_1D" "-i" "${dir}/${file}"
```
jbrillon, line 66: In `$1/PHiLiP_...`, is `1` a variable holding the directory path? And where is it defined?

jbrillon (commented):
Good work, see comments below

src/testing/reduced_order_pod_adaptation.h
donovan97: No, there's no need. It's just an empty constructor.

tests/integration_tests_control_files/reduced_order/pod_adaptation_snapshot_generator.sh
donovan97: In Bash, $1 is the first argument given when calling the script. So in my cmake I have ./pod_adaptation_snapshot_generator.sh ${EXECUTABLE_OUTPUT_PATH}; so $1 refers to ${EXECUTABLE_OUTPUT_PATH}.

src/testing/reduced_order_pod_adaptation.h
donovan97: It will take the default template arguments.

tests/integration_tests_control_files/reduced_order/pod_adaptation_snapshot_generator.sh
jbrillon: Ah ok, I had forgotten about this, thanks for explaining that!

src/testing/reduced_order_pod_adaptation.h
```diff
@@ -0,0 +1,70 @@
...
+    /// Constructor.
+    ReducedOrderPODAdaptation(const Parameters::AllParameters *const parameters_input);
+
+    /// Run test
+    int run_test () const override;
+};
+
+///Functional to take the integral of the solution
+template <int dim, int nstate, typename real>
+class BurgersRewienskiFunctional : public Functional<dim, nstate, real>
```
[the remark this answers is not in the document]
donovan97: It could be used for other equations, but it just takes the integral of the solution, so it's not very interesting. Since it's not used anywhere else, I figured it might be best to put it in the test not to clutter the functional directory with a functional that's only used once.

donovan97:
I have found a bug while running the tests on Narval. the /usr/bin/mpirun command needs to be changed to mpirun only. All required tests are now passing.

src/testing/reduced_order_pod_adaptation.h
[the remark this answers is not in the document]
PranshulThakur: Sure, if it's only used for this test.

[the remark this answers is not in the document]
PranshulThakur: Sure, if it's only used for this test.

PranshulThakur (approved):

PR #95, opened 2022-01-03
Mesh h-Adaptation Part 1

PranshulThakur:
The goal of this commit is to move the current hard-coded residual mesh adaptation to a separate class. Future goals include introducing goal oriented mesh adaptation using different error estimators in subsequent pull requests.

Test on transonic Gaussian bump with a shock will be implemented to test h-adaptation near the shock.


src/mesh/high_order_grid.cpp
```diff
@@ -1050,6 +1050,24 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::execute_coarsen
     //}
 }
 
+template <int dim, typename real, typename MeshType, typename VectorType, typename DoFHandlerType>
+dealii::Point<dim> HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::smallest_cell_coordinates()
+{
+    dealii::Point<dim> smallest_cell_coord;
+    double min_diameter = triangulation->begin_active()->diameter();
```
dougshidong, line 1057: The cell->diameter() returns the Triangulation straight-sided cell's diameter. When we do shape optimization, we end up moving the high-order nodes, while the Triangulation's cell vertices stay at the same place! In fact, the Triangulation cells simply becomes a connectivity map rather than anything useful physically.

Can you clarify what definition of "diameter" that you need for hexahedra?

src/mesh/high_order_grid.cpp
```diff
@@ -1050,6 +1050,24 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::execute_coarsen
...
+    for (const auto &cell : triangulation->active_cell_iterators()) 
+    {
+        if (min_diameter > cell->diameter())
+        {
+            min_diameter = cell->diameter();
+            smallest_cell_coord = cell->vertex(0);
```
dougshidong, line 1063: Any reason you're just querying the first cell's vertex as the cell's location?

src/mesh/high_order_grid.cpp
dougshidong: Nevermind, I now understand that it's just meant for the test.

src/mesh/high_order_grid.cpp
dougshidong: Nevermind, I now understand that it's just meant for the test.

src/mesh/high_order_grid.cpp
dougshidong: Worth keeping in mind though that it's not the actual cell sizes in more complex cases.

src/mesh/mesh_adaptation.cpp
```diff
@@ -0,0 +1,119 @@
...
+MeshAdaptation<dim,real,MeshType>::MeshAdaptation(double critical_res_input, int total_ref_cycle, double refine_frac, double coarsen_frac)
+    : critical_residual(critical_res_input)
+    , total_refinement_cycles(total_ref_cycle)
+    , refinement_fraction(refine_frac)
+    , coarsening_fraction(coarsen_frac)
+    , current_refinement_cycle(0)
+    {}
+
+template <int dim, typename real, typename MeshType>
+int MeshAdaptation<dim,real,MeshType>::adapt_mesh(std::shared_ptr< DGBase<dim, real, MeshType> > dg)
```
dougshidong, line 15: Instead of returning 0s, make the function void!

dougshidong: Same applies for the other functions. If you don't use the return value, there is no point in returning it.

src/mesh/mesh_adaptation.cpp
```diff
@@ -0,0 +1,119 @@
...
+        return 0;
+    }
+
+    cellwise_errors.reinit(0);
+    cellwise_errors.reinit(dg->high_order_grid->triangulation->n_active_cells());
+    compute_cellwise_errors(dg);
+
+    fixed_fraction_isotropic_refinement_and_coarsening(dg);
+    current_refinement_cycle++;
+    std::cout<<"Refined"<<std::endl;
```
dougshidong, line 33: This will get outputted by every processor. Look into the code for other uses of pcout where only the first processor outputs this message.

src/mesh/mesh_adaptation.cpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    if (total_refinement_cycles == current_refinement_cycle)
+    {
+        return 0;
+    }
+
+    cellwise_errors.reinit(0);
+    cellwise_errors.reinit(dg->high_order_grid->triangulation->n_active_cells());
+    compute_cellwise_errors(dg);
```
dougshidong, line 29: Keep in mind the Single Responsibility Principle. Currently, your class is doing 2 big things:
1. Computing the error estimates
2. Refining
Consider separating the error estimate into a separate class whose single responsibility is to compute errors.
Then pass this error estimate to this mesh adaptation class whose single responsibility is to refine our DG class.

src/mesh/mesh_adaptation.cpp
```diff
@@ -0,0 +1,119 @@
...
+         const dealii::FESystem<dim,dim> &fe_ref = dg->fe_collection[i_fele];
+         const unsigned int n_dofs_cell = fe_ref.n_dofs_per_cell();
+         dofs_indices.resize(n_dofs_cell);
+         cell->get_dof_indices (dofs_indices);
+         double max_residual = 0;
+         for (unsigned int idof = 0; idof < n_dofs_cell; ++idof) 
+         {
+             const unsigned int index = dofs_indices[idof];
+             const unsigned int istate = fe_ref.system_to_component_index(idof).first;
+             if (istate == dim+2-1)
```
dougshidong, line 65: This won't work for equations other than Euler/NS. Any reason why you check for that istate specifically?

dougshidong (commented):
Looks pretty good! Consider decoupling the error computation and refinement.

src/mesh/high_order_grid.cpp
PranshulThakur: Thanks, will keep it in mind (and I have introduced mapping to get cell vertex).

src/mesh/high_order_grid.cpp
PranshulThakur: Yeah, it's just to get a ball park estimate of the refinement location.

src/mesh/mesh_adaptation.cpp
PranshulThakur: Sure, fixed it.

src/mesh/mesh_adaptation.cpp
PranshulThakur: Thanks, it now uses pcout.

src/mesh/mesh_adaptation.cpp
PranshulThakur: Nope, thanks for pointing it out. That check is removed.

src/mesh/high_order_grid.cpp
```diff
@@ -1050,6 +1050,49 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::execute_coarsen
...
+        current_cell_diameter = cell->diameter();
+        if ((min_diameter_local > current_cell_diameter) && (cell->is_locally_owned()))
+        {
+            min_diameter_local = current_cell_diameter;
+            smallest_cell_coord = mapping_fe_field->transform_unit_to_real_cell(cell, unit_vertex);
+            //smallest_cell_coord = cell->center();
+        }
+    }
+    
+    MPI_Barrier(mpi_communicator); // wait till all processors are done
```
dougshidong, line 1075: No need to wait until they're all done. MinMaxAvg is a collective operation that will wait anyways.

src/mesh/high_order_grid.cpp
```diff
@@ -1050,6 +1050,49 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::execute_coarsen
...
+
+    if (iproc == n_proc_small)
+    {
+        smallest_cell_global = smallest_cell_coord;         
+    }
+    
+    MPI_Barrier(mpi_communicator); // wait for all processors
+
+    for (int i=0; i<dim; i++)
+        smallest_cell_global[i] = dealii::Utilities::MPI::sum(smallest_cell_global[i],mpi_communicator); // Update values in all processors
```
dougshidong, line 1090: Look into using MPI_Bcast. While sum might work, it's not the correct operation that tells the reader of the code what you're doing. Relevant tutorials:
https://mpitutorial.com/tutorials/mpi-broadcast-and-collective-communication/
https://mpitutorial.com/tutorials/mpi-scatter-gather-and-allgather/

src/testing/euler_naca0012_optimization.cpp
```diff
@@ -550,7 +550,7 @@ ::optimize (const unsigned int nx_ffd, const unsigned int poly_degree) const
     dealii::VectorTools::interpolate(dg->dof_handler, initial_conditions, dg->solution);
     // Create ODE solver and ramp up the solution from p0
     std::shared_ptr<ODE::ODESolverBase<dim, double>> ode_solver = ODE::ODESolverFactory<dim, double>::create_ODESolver(dg);
-    ode_solver->n_refine = 0;
+    //ode_solver->n_refine = 0;
```
dougshidong, line 553: Instead of commenting it out. Just delete. Git saves this information.

PranshulThakur: Deleted it.

src/mesh/high_order_grid.cpp
PranshulThakur: Thanks for the links. This function now uses MPI_Bcast.

src/mesh/high_order_grid.cpp
PranshulThakur: Thanks, deleted that line.

src/mesh/high_order_grid.cpp
```diff
@@ -1071,24 +1071,26 @@ dealii::Point<dim> HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::s
...
+    double global_point[dim];
 
     if (iproc == n_proc_small)
     {
-        smallest_cell_global = smallest_cell_coord;         
+       for (int i=0; i<dim; i++)
+            global_point[i] = smallest_cell_coord[i];
     }
     
     MPI_Barrier(mpi_communicator); // wait for all processors
```
dougshidong, line 1086: No need for the Barrier. You almost never need MPI_Barrier unless you're debugging or sending non-blocking information

src/mesh/mesh_adaptation.cpp
```diff
@@ -0,0 +1,117 @@
...
+         {
+            const unsigned int index = dofs_indices[idof];
+            const double res = std::abs(dg->right_hand_side[index]);
+            if (res > max_residual) 
+                max_residual = res;
+         }
+         cellwise_errors[cell->active_cell_index()] = max_residual;
+     }
+
+    return;
```
dougshidong, line 72: Unnecessary "return;" at the end of a void function. Same applies for some of the other functions you provided.

src/mesh/mesh_adaptation.cpp
```diff
@@ -0,0 +1,117 @@
...
+    , refinement_fraction(refine_frac)
+    , coarsening_fraction(coarsen_frac)
+    , current_refinement_cycle(0)
+    , pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)==0)
+    {}
+
+template <int dim, typename real, typename MeshType>
+void MeshAdaptation<dim,real,MeshType>::adapt_mesh(std::shared_ptr< DGBase<dim, real, MeshType> > dg)
+{
+    if (!refine_mesh)
```
donovan97, line 18: Maybe this check should be done in ode_solver_base for clarity, since the value of refine_mesh seems to be determined in ode_solver_base?

src/ode_solver/ode_solver_base.cpp
```diff
@@ -148,10 +151,10 @@ int ODESolverBase<dim,real,MeshType>::steady_state ()
                 this->dg->output_results_vtk(file_number);
             }
         }
-
-        if ( refine && this->residual_norm < 1e-9 && i_refine < n_refine) {
-            i_refine++;
-            dg->refine_residual_based();
+        
+        if (this->residual_norm < meshadaptation->critical_residual)
```
donovan97, line 155: The refine_mesh flag seems to only get initialized in the initialize_steady_polynomial_ramping function. What if the steady_state function is called directly?

src/mesh/mesh_adaptation.cpp
PranshulThakur: Error estimation is now done by a separate class and has a general virtual function for future extension.

src/mesh/high_order_grid.cpp
PranshulThakur: Deleted it.

src/mesh/high_order_grid.cpp
```diff
@@ -1050,6 +1050,49 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::execute_coarsen
...
 
+template <int dim, typename real, typename MeshType, typename VectorType, typename DoFHandlerType>
+dealii::Point<dim> HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::smallest_cell_coordinates()
+{
+    const int iproc = dealii::Utilities::MPI::this_mpi_process(mpi_communicator);
+    const dealii::Point<dim> unit_vertex = dealii::GeometryInfo<dim>::unit_cell_vertex(0);
+    double min_diameter_local;
+    double current_cell_diameter;
+    auto cell = dof_handler_grid.begin_active();
+    auto endcell = dof_handler_grid.end();
```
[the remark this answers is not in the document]
PranshulThakur: They are all in the loop now.

src/mesh/mesh_adaptation.cpp
PranshulThakur: Removed return at the end.

src/mesh/mesh_adaptation.cpp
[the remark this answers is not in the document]
PranshulThakur: In the new commit, adapt_mesh() just adapts the meshes (without any checks) and, since the refine_mesh check is only for the ODE solver (and will not be required outside of that class), it's put back in ODEsolver as mesh_refinement_in_ode_solver.

PranshulThakur: Yes, it's in ODESolver now.

src/mesh/mesh_adaptation.cpp
```diff
@@ -0,0 +1,117 @@
...
+    {
+        return;
+    }
+
+    if (total_refinement_cycles == current_refinement_cycle)
```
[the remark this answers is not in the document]
PranshulThakur: This check is now performed in ODESolver.

Since adapt_mesh() adapts meshes just once, some sort of loop must exist outside wherever it's called for multiple refinements. Members `total_refinement_cycles` and `current_refinement_cycles` are still members of MeshAdaptation, but they are not used by the member functions and simply exist for book-keeping purposes (to make it easy for someone calling the mesh_adaptation object in a loop to set conditions to exit the loop).

src/ode_solver/ode_solver_base.cpp
PranshulThakur: In the new commit, refine_mesh_in_ode_solver is set `true` to begin with.

If someone calls steady_state() directly and does not want mesh adaptation, the default parameter values of total_refinement_cycles and current_refinement_cycle are set to 0 and adapt_mesh() will not be called.

If someone wants to use steady_state() with mesh_adaptation, they can modify the values of total_refinement_cycles in the parameter file.

src/mesh/high_order_grid.cpp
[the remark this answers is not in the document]
dougshidong: I love the fact that you're using range-based for loops. It's the "best" way to do it now.

dougshidong:
All the comments on my side have been addressed. I will let @donovan97 resolve his comments and merge the pull request in the `master` branch when he is satisfied. Thank you @PranshulThakur for resolving those comments and @donovan97 for the feedback.

One final note, @PranshulThakur could you confirm that this branch fully compiles and passes all the necessary tests?

PranshulThakur:
Sure, ctest is running. I will send an update soon.

PranshulThakur:
Yes, all tests have passed (except the 5 stated in issues).

PR #97, opened 2022-01-04
Cluster Compile Scripts

jbrillon:
For compiling PHiLiP on the Compute Canada Narval cluster, includes a generalized job compile script and some clean up to the comments in the scripts.


donovan97 (commented):
Looks good to me.

PR #105, opened 2022-02-08
Explicit ODE Solver Restructure

cpethrick:
Restructuring `explicit_ode_solver` to use a Butcher tableau for computing Runge-Kutta time-stepping rather than hard-coded. Added a test to check the order of time-stepping.


src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -13,90 +13,43 @@ void ExplicitODESolver<dim,real,MeshType>::step_in_time (real dt, const bool pse
...
+    for (int i = 0; i < rk_order; ++i){
+        this->rk_stage[i] = this -> solution_update; //u_n
+        for (int j = 0; j < i; ++j){
+            if (this->butcher_a[i][j] != 0){
+                if (pseudotime) {
+                    //implemented but not tested 
+                    //to my knowledge, there aren't any existing tests using explicit steady-state 
+                    //(searched through unit_tests folder)
+                    std::cout << "Explicit pseudotime not tested!!" << std::endl;
+                    const double CFL =this-> butcher_a[i][j] * dt;
```
donovan97, line 32:
```suggestion
                    const double CFL =this->butcher_a[i][j] * dt;
```

donovan97 (approved):
Looks good to me.

src/ode_solver/implicit_ode_solver.cpp
```diff
@@ -1,5 +1,6 @@
 #include "implicit_ode_solver.h"
```
dougshidong, line 1: Since I see that the changes are trivial, this file is a sign that you added all the files at some point through:
git add . or git add -A or git add -u

My suggestion is to add relevant files manually with the purpose of the commit description in mind.

Fine for this PR, but careful in the future as it might add a lot of files you don't want it to.

tests/unit_tests/ode_solver_unit_test/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+      COMMAND mpirun -n 1 ${EXECUTABLE_OUTPUT_PATH}/${TEST_TARGET}
+      WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
+    )
+
+    unset(dim)
+    unset(TEST_TARGET)
+    unset(PhysicsLib)
+    unset(NumericalFluxLib)
+    unset(ODESolverLib)
+    unset(ParameterLib)
```
dougshidong, line 39: No need to unset PhysicsLib, NumericalFluxLib, and ParameterLib since they weren't set.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,217 @@
...
+#endif
+
+            //Advects a sine wave (1D) with peridic BCs and compares to anlytical solution
+            //refines spatial discretization 5 times and writes a convergence table to file
+            //Can be used to verify spatial order of the explicit ODE solver
+            //Note: 2D is implemented but is not included in CMakeLists.txt (untested)
+
+
+            typename dealii::Triangulation<dim>::MeshSmoothing(
+                dealii::Triangulation<dim>::smoothing_on_refinement |
```
dougshidong, line 42: Odd indenting.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,217 @@
...
+    conv_tab_file.close();
+
+    for (int i = 0; i < n_time_refinements; ++i){
+        if (abs(L2_error_conv_rate[i] - expected_order) > order_tolerance){
+            testfail = 1;
+            std::cout << "Expected convergence order was not reached at refinement " << i + 1 <<std::endl;
+        }
+    }
+
+    return testfail; //always passes
```
dougshidong, line 215: Remove comment about it always passing.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,217 @@
...
+            //Note: 2D is implemented but is not included in CMakeLists.txt (untested)
+
+
+            typename dealii::Triangulation<dim>::MeshSmoothing(
+                dealii::Triangulation<dim>::smoothing_on_refinement |
+                dealii::Triangulation<dim>::smoothing_on_coarsening));
+
+    int testfail = 0;
+
+    double left = 0.0;
```
dougshidong, line 47: As a rule of thumb, favour constness throughout the code.

Basically, const everything except when you can't. Please go through the PR and change non-const variable to const (when it makes sense).

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,217 @@
...
+                expression_exact,
+                constants);
+
+        //Calculating L2 error
+        dealii::Vector<double> difference_per_cell(grid->n_active_cells());
+        dealii::VectorTools::integrate_difference(dg->dof_handler, 
+                dg->solution, 
+                exact_solution, 
+                difference_per_cell, 
+                dealii::QGauss<dim>(space_poly_degree+1), //check that this is correct polynomial degree
```
dougshidong, line 166: Maybe you want to overintegrate? With a p+1 integration, you might have some error in your error calculation.

I would try out something like p+5 or p+10 just to see if you get better/worse convergence rates.

dougshidong (commented):
Very cool PR. Much cleaner explicit ODE solver structure with extensibility in mind.

Excited to see upcoming work.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,217 @@
...
+    double dt = dt_init;
+    const double refine_ratio = 0.5;
+    double finalTime = 1.0;
+
+    dealii::ConvergenceTable convergence_table;
+
+    double L2_error_old = 0;
+    double L2_error_conv_rate[4];
+
+    for (int refinement = 0; refinement < n_time_refinements+1; ++refinement){
```
PranshulThakur, line 103: In `refinement < n_time_refinements+1` , the +1 is probably not required. Setting `n_time_refinements` to 4 should work too.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,217 @@
...
+    int n_time_refinements = 3;
+    double dt_init = 2.5E-3;
+    double dt = dt_init;
+    const double refine_ratio = 0.5;
+    double finalTime = 1.0;
+
+    dealii::ConvergenceTable convergence_table;
+
+    double L2_error_old = 0;
+    double L2_error_conv_rate[4];
```
PranshulThakur, line 101: The size of `L2_error_conv_rate` is hard-coded. Maybe set it to `n_time_refinements-1`? And to be consistent, you can also use `dealii::Vector<double>` for this.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,217 @@
...
+    double left = 0.0;
+    double right = 2.0;
+    const bool colorize = true;
+    int n_refinements = 5;
+
+    //generating grid
+    dealii::GridGenerator::hyper_cube(*grid, left, right, colorize);
+    //setting periodic BC 
+    std::vector<dealii::GridTools::PeriodicFacePair<typename Triangulation::cell_iterator> > matched_pairs;
+    if (dim == 2){
```
PranshulThakur, line 56: Since this test is only for dim = 1 or 2, it's better to give an error message if dim !=1,2.

PranshulThakur (commented):
Great work! I have added some minor comments.

src/ode_solver/implicit_ode_solver.cpp
cpethrick: Yes, this was a mistake in the first commit of the branch. I've been more careful with adding files in subsequent commits!

src/ode_solver/explicit_ode_solver.cpp
cpethrick: Thanks for the formatting suggestions where I've been inconsistent!

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
cpethrick: There is only a small change when overintegrating, e.g. 1.2667e-10 with p+1 to 1.2890e-10 with p+10, I think because the sine wave initial condition yields a smooth enough solution. I will increase the order as you've recommended for a more robust error calculation if another user uses a less "nice" initial condition. Thanks!

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -13,90 +13,43 @@ void ExplicitODESolver<dim,real,MeshType>::step_in_time (real dt, const bool pse
...
+    this->solution_update = this->dg->solution; //u_ni
+    for (int i = 0; i < rk_order; ++i){
+        this->rk_stage[i] = this -> solution_update; //u_n
+        for (int j = 0; j < i; ++j){
+            if (this->butcher_a[i][j] != 0){
+                if (pseudotime) {
+                    //implemented but not tested 
+                    //to my knowledge, there aren't any existing tests using explicit steady-state 
+                    //(searched through unit_tests folder)
+                    std::cout << "Explicit pseudotime not tested!!" << std::endl;
```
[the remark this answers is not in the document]
cpethrick: Thanks, I didn't know about that!

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
cpethrick: You're right, I was doing this in a confusing way - I was viewing only calculations after the first ones as "refinement". I'll follow your suggestion so that it is less confusing.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
cpethrick: Good suggestion, `dealii::Vector<double>` is a better way to handle this.

dougshidong: Every time you use a C-style array like this, consider using either a C++ std::vector by default or std::array if performance matters.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
cpethrick: Thanks, I'll add an error check for `dim`.

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -13,90 +13,43 @@ void ExplicitODESolver<dim,real,MeshType>::step_in_time (real dt, const bool pse
...
-            this->pcout<< "done." << std::endl;
-        }
+   
+    
+    //calculating stages
+    this->solution_update = this->dg->solution; //u_ni
+    for (int i = 0; i < rk_order; ++i){
+        this->rk_stage[i] = this->solution_update; //u_n
+        for (int j = 0; j < i; ++j){
+            if (this->butcher_a[i][j] != 0){
```
jbrillon, line 26:
```suggestion
            if (this->butcher_a[i][j] != 0.0){
```

jbrillon: Although it doesn't actually make a difference... just a preference since `butcher_a` is a non `int`

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -13,90 +13,43 @@ void ExplicitODESolver<dim,real,MeshType>::step_in_time (real dt, const bool pse
...
+        this->dg->assemble_residual (); // RHS : du/dt = RHS = F(u_n + dt* sum(a_ij*k_j)
+        this->dg->global_inverse_mass_matrix.vmult(this->rk_stage[i], this->dg->right_hand_side); //rk_stage[i] = IMM*RHS = F(u_n + dt*sum(a_ij*k_j)
     }
-    else {
-        this->pcout << "Invalid runge_kutta_order." << std::endl;
-        std::abort();
+
+    //assemble solution from stages
+    for (int i = 0; i < rk_order; ++i){
+        this->solution_update.add(dt* this->butcher_b[i],this->rk_stage[i]);
```
jbrillon, line 47: Since this is technically the last stage of the RK scheme, could we replace `butcher_b[i]` by `runge_kutta_coefficent_matrix[i][rk_order]` where `runge_kutta_coefficent_matrix` has size `(rk_order,rk_order+1)`

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -13,90 +13,43 @@ void ExplicitODESolver<dim,real,MeshType>::step_in_time (real dt, const bool pse
...
+                    const double CFL =this->butcher_a[i][j] * dt;
+                    this->dg->time_scale_solution_update(this->rk_stage[j], CFL);
+                    this->rk_stage[i].add(1.0,  this->rk_stage[j]);
+                } else {
+                    this->rk_stage[i].add(dt*this->butcher_a[i][j], this->rk_stage[j]);
+                }
+            }
+        } //u_n + dt* sum(a_ij *k_j)
+        this->dg->solution = this->rk_stage[i];
+        this->dg->assemble_residual (); // RHS : du/dt = RHS = F(u_n + dt* sum(a_ij*k_j)
```
jbrillon, line 41:
```suggestion
        this->dg->assemble_residual(); // RHS : du/dt = RHS = F(u_n + dt* sum(a_ij*k_j)
```

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -107,10 +60,39 @@ void ExplicitODESolver<dim,real,MeshType>::allocate_ode_system ()
...
+    for (int i=0; i<rk_order; i++) {
         this->rk_stage[i].reinit(this->dg->solution);
     }
+
+    //Assigning butcher tableau
+    this->butcher_a.reinit(rk_order,rk_order);
+    this->butcher_b.reinit(rk_order);
+    if (rk_order == 3){
+        //RKSSP3
+        const double butcher_a_values[9] = {0,0,0,1.0,0,0,0.25, 0.25, 0};
```
jbrillon, line 75: `butcher_a_values` seems to always require the first row to be zeros, if that is always the case in the future, is there a way you could modify the implementation to avoid having to specify these? for example, this line would simply be `const double butcher_a_values[9] = {1.0,0,0,0.25, 0.25, 0};`

jbrillon: and if you use the coefficient matrix idea I mentioned earlier, it would be `const double runge_kutta_coefficients[rk_order*rk_order] = {1.0,0,0, 0.25,0.25,0, 1.0/6.0, 1.0/6.0, 2.0/3.0};` then these would fill the `runge_kutta_coefficient_matrix`

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    std::shared_ptr<Triangulation> grid = std::make_shared<Triangulation>(
+#if PHILIP_DIM!=1
+            MPI_COMM_WORLD,
+#endif
+
+    //Advects a sine wave (1D) with peridic BCs and compares to anlytical solution
+    //refines spatial discretization 5 times and writes a convergence table to file
+    //Can be used to verify spatial order of the explicit ODE solver
+    //Note: 2D is implemented but is not included in CMakeLists.txt (untested)
```
jbrillon, line 43: You're adding the description comment within the definition of the `grid` variable, try to avoid this since we can easily forget that you're between two brackets `(` (open - line 35) and `)` (close - line 48)

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "dg/dg_factory.hpp"
+#include "ode_solver/ode_solver_factory.h"
+
+
+int main (int argc, char * argv[])
+{
+    dealii::Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);
+    const int dim = PHILIP_DIM;
+    
+    if ( (dim != 1) && (dim != 2) ){
```
jbrillon, line 24:
```suggestion
    if ((dim != 1) && (dim != 2)){
```

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,220 @@
...
+        }
+        else if (dim == 1){
+            variables = "x";
+            //expression_initial = "exp(- 20 * (x-1) * (x-1))";	
+            expression_initial = "sin(2*pi*x/2.0)";	
+        }
+        initial_condition.initialize(variables,
+                expression_initial,
+                constants);
+        dealii::VectorTools::interpolate(dg->dof_handler,initial_condition,dg->solution);
```
jbrillon, line 139: your initial condition can be added inside `src/physics/initial_conditions/initial_condition.cpp`, also take a look at somehow using `src/testing/flow_solver.cpp` to take care of all these standard tasks (create ODE, create DG, create grid, initialize solution etc) including advancing the solution in time, @donovan97 has used the flow solver within at least one of his newly implemented test cases for doing that

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,220 @@
...
+    dealii::Vector<double> L2_error_conv_rate(n_time_calculations);
+
+    for (int refinement = 0; refinement < n_time_calculations; ++refinement){
+
+        //initial_time_step is not modified by explicit ODE solver
+        all_parameters.ode_solver_param.initial_time_step = dt;
+        std::cout << "Using time step = " << dt << std::endl;
+        std::cout << "refinement = " << refinement << std::endl;
+
+        all_parameters.ode_solver_param.output_solution_every_x_steps = int(finalTime/dt/10.0); //output 10 vtk files (if dt reaches finalTime exactly)
```
jbrillon, line 114: there's also a parameter, `output_solution_every_dt_time_intervals`, that you could take advantage of which would be more intuitive, replacing this line by: `all_parameters.ode_solver_param.output_solution_every_dt_time_intervals = int(finalTime/10.0)`

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
```diff
@@ -0,0 +1,220 @@
...
+        ode_solver->advance_solution_time(finalTime);
+
+        //Defining exact solution
+        dealii::FunctionParser<dim> exact_solution;
+        constants["a_x"] = adv_speed_x;
+        constants["a_y"] = adv_speed_y;
+        constants["t"] = ode_solver -> current_time;//finalTime;
+        std::string expression_exact;
+        if (dim == 2){
+            expression_exact = "exp( -( 20*(x-1-a_x*t)*(x-1-a_x*t) + 20*(y-1-a_y*t)*(y-1-a_y*t) ) )";
```
jbrillon, line 154: If these are the exact solutions to the unsteady problem you're simulating and checking the order with, maybe we should have some sort of `ExactSolutionFactory` similar to `InitialConditionFactory`, where you'd create the directory `src/physics/exact_solutions` and create the files in a similar fashion -- this will be good for when we want to check the temporal order of convergence for different problems

jbrillon (commented):
Great first PR! I've left some comments that would change the implementation of your test quite a lot but overall would provide a sustainable infrastructure for doing temporal convergence studies. Last comment, consider moving/renaming `tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp` to `src/testing/temporal_study.cpp` -- this may be more appropriate since it is similar to `grid_study.cpp`

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
cpethrick: Thanks for pointing that out, I will move that comment.

tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp
cpethrick: Thanks for the suggestion, I will change this.

src/ode_solver/explicit_ode_solver.cpp
cpethrick: Changed, thanks!

PranshulThakur (approved):

cpethrick:
I have made changes according to all comments on formatting, naming etc. I will continue to work on the more labour-intensive suggestions (i.e., moving the grid generation to a more general location; reformulating the current test case to use flow_solver; moving initial conditions to the more general location; and creating an exact solution in the same way as initial conditions).

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -107,10 +60,39 @@ void ExplicitODESolver<dim,real,MeshType>::allocate_ode_system ()
...
+    for (int i=0; i<rk_order; i++) {
         this->rk_stage[i].reinit(this->dg->solution);
     }
+
+    //Assigning butcher tableau
+    this->butcher_tableau_a.reinit(rk_order,rk_order);
+    this->butcher_tableau_b.reinit(rk_order);
+    if (rk_order == 3){
+        //RKSSP3
+        const double butcher_tableau_a_values[9] = {0,0,0,1.0,0,0,0.25, 0.25, 0};
```
jbrillon, line 75:
```suggestion
        const double butcher_tableau_a_values[9] = {0,0,0,1.0,0,0,0.25,0.25,0};
```

cpethrick:
I have made changes according to all comments on formatting, naming etc. I will continue to work on the more labour-intensive suggestions in a future PR (i.e., moving the grid generation to a more general location; reformulating the current test case to use flow_solver; moving initial conditions to the more general location; and creating an exact solution in the same way as initial conditions). Once this PR is approved, I will open an issue for these changes and assign it to myself.

jbrillon:
Sounds good, can you confirm that all the tests that are expected to pass are indeed passing?

dougshidong:
Please go ahead and open up the issues you'd assign to yourself. That way the reviewers know that their comments will be taken care by specific issues in future PRs. 

Also, let us know if the tests pass on your end.

cpethrick:
Thanks for the suggestion to run all of the tests. Unfortunately, one of the integration tests now fails (MPI_2D_ADVECTION_EXPLICIT_MANUFACTURED_SOLUTION_LONG) because the explicit pseudotime implementation is not correct. I'm currently trying to debug this, as it should be fixed before this PR is added to the code.

Doug, I will open those issues now.

cpethrick:
I have fixed the issue which was causing MPI_2D_ADVECTION_EXPLICIT_MANUFACTURED_SOLUTION_LONG to fail. Results now match to the convergence tolerance for explicit pseudotimestepping. Removed an extra call of `assemble_residual()` which was causing a performance issue, and run-times are similar between master and this branch.

dougshidong:
Looks good to me.

Thanks again for the great PR and making all those changes.

PR #107, opened 2022-02-17
NACA0012 Mesh Files Intall Instructions

jbrillon:
This pull request is to update the installation instructions for the NACA0012 mesh files that must be downloaded externally, and provides a convenient solution for the cluster; see files changed.


dougshidong (approved):
Lgtm

PR #108, opened 2022-02-18
Bash Script for Downloading NACA0012 Mesh Files

### LINKED ISSUE #94: Automate download of NACA mesh files from GoogleDrive in bash install script
To avoid having to always manually download and copy these files. Can use `gdown`, see [stack overflow issue - wget/curl large file from google drive](https://stackoverflow.com/questions/25010369/wget-curl-large-file-from-google-drive)

jbrillon:
This pull request closes #94 by automating the download of the mesh files from GoogleDrive using [gdown](https://github.com/wkentaro/gdown) -- this is intended to be used for installing on local machines only (i.e. not on the cluster as this has already been address there).


donovan97 (approved):
Looks good.

PR #117, opened 2022-03-31
Fix gmsh reader 2D cell rotation

dougshidong:
Fix issue brought up by @MayaTatarelli not being able to read .msh file.

Attached is the .msh file, the .geo used to generate it (requires Gmsh with OpenCascade), and the resulting output when run with the previous code.

More notatbly, the output "Couldn't find rotation..." shows that we are not able to read the element vertices due to improper rotations

[airfoil.zip](https://github.com/dougshidong/PHiLiP/files/8386833/airfoil.zip)

1. This issue has been fixed by including the additional cell vertices permutation (left-hand rule ordering or flipping the Z-axis direction).
2. The .msh that exposed the issue has been added to the unit test suite.

@kit71717 Calvin, you might want to review this PR since you are likely doing something similar in 3D.

PR #119, opened 2022-03-31
Integration test MPI_VISCOUS_TAYLOR_GREEN_VORTEX_ENERGY_CHECK

jbrillon:
Pull request to incorporate the addition of a integration test for viscous Taylor Green Vortex (TGV) which computes the kinetic energy at some final time, using the `flow_solver` to setup and advance the solution, and compares the resulting value with an expected value, checking that the relative error is within some tolerance. Integration tests added:
- `MPI_VISCOUS_TAYLOR_GREEN_VORTEX_ENERGY_CHECK_QUICK`
- `MPI_VISCOUS_TAYLOR_GREEN_VORTEX_ENERGY_CHECK_LONG`

Additional contributions:
- Updating `get_NACA0012_mesh_files_cluster.sh` to include `airfoil.msh` in regards to PR #117


jbrillon:
All added integration tests passing @cpethrick

src/testing/flow_solver.cpp
```diff
@@ -96,6 +96,12 @@ int FlowSolver<dim,nstate>::run_test() const
         flow_solver_case->steady_state_postprocessing(dg);
     }
     pcout << "done." << std::endl;
+}
+
+template <int dim, int nstate>
+int FlowSolver<dim,nstate>::run_test() const
```
donovan97, line 102: Is there a reason for creating run_flow_solver instead of just having it in run_test?

src/testing/flow_solver_cases/periodic_cube_flow.h
```diff
@@ -30,6 +30,9 @@ class PeriodicCubeFlow : public FlowSolverCaseBase<dim,nstate>
     /// Destructor
     ~PeriodicCubeFlow() {};
 
+    /// Integrates over the entire domain
+    double integrate_over_domain(DGBase<dim, double> &dg,const std::string integrate_what) const;
```
donovan97, line 34: integrate_what could be an enum instead of a string.

src/testing/flow_solver.cpp
jbrillon: Yeah, since run test returns an `int`, having a `void` function lets us avoid that

src/testing/flow_solver_cases/periodic_cube_flow.h
PranshulThakur, line 34: To be consistent with other functions in this class, maybe `integrate_over_domain()` can take a shared_ptr of DGBase instead of a reference.

PranshulThakur (commented):
I have added one comment, but the PR looks great!

src/testing/flow_solver_cases/periodic_cube_flow.h
dougshidong: No need. Only pass shared_ptr if the goal is to store it long term to be reused. In this case, it's a function, so passing by reference is the correct choice.

src/testing/flow_solver.cpp
donovan97: Its a bit confusing to have an extra function just to avoid returning a value. If you just use the run_test function and don't use the return value, does the compiler complain? If it does you could try static_cast<void>(run_test()).

dougshidong (approved):
Looks good once a resolution has been reached with other reviewers.

src/testing/flow_solver.cpp
donovan97: I think eventually `FlowSolver` should be given its own directory and no longer be in /src/testing, since its not really a test. In that case, `run_flow_solver()` does make more sense. But for now, to run it outside of the class, instead of `int success = flow_solver.run_test()`, you could try `static_cast<void>(run_test());`, I think it'll eliminate the need to pass the return value to a variable (not sure I haven't tried it). Otherwise, I guess you can just leave it as it is, and we can change it someday if we pull `FlowSolver` out of `/src/testing`.

PranshulThakur (approved):
Looks good to me.

src/testing/flow_solver.cpp
dougshidong: Agreeing with Donovan here. The flow solver would be moved out at some point to its own directory and it would return a more significant enum regarding convergence. For now the scope of this PR is to add the integration test. Simply open an issue to move the flow solver, reference this conversation, and resolve this conversation.

Regarding the suppression of warning error: casting it C-style as `(void)run_test();` is what I have been doing and used more commonly for unused variables.

jbrillon: Addressed in most recent commit

src/testing/flow_solver_cases/periodic_cube_flow.h
jbrillon: Addressed in most recent commit

src/parameters/parameters_flow_solver.cpp
```diff
@@ -43,6 +43,14 @@ void FlowSolverParam::declare_parameters(dealii::ParameterHandler &prm)
         prm.declare_entry("sensitivity_table_filename", "sensitivity_table",
                           dealii::Patterns::FileName(dealii::Patterns::FileName::FileType::input),
                           "Filename for the sensitivity data table output file: sensitivity_table_filename.txt.");
+
+        prm.enter_subsection("taylor_green_vortex_energy_check");
+        {
+            prm.declare_entry("expected_kinetic_energy_at_final_time", "1",
+                              dealii::Patterns::Double(1e-15, 10000000),
```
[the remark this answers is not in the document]
jbrillon: Addressed in most recent commit

donovan97 (approved):

PR #122, opened 2022-04-16
Restart Flow Solver

### LINKED ISSUE #102: FlowSolver restart
Implement the ability to restart a computation from a particular saved iteration in the `FlowSolver` class. This is needed for running large cases on the cluster.

jbrillon:
Pull request to close #102 

Can now restart the `FlowSolver` from a specified restart file index. This includes:
- Initializing the `dg->solution` from files written using `dealii::SolutionTransfer`
- Initializing the `unsteady_data_table`, a `dealii::TableHandler` object, from a file with the added reader `initialize_data_table_from_file()`

The implementation conveniently dumps an associated parameters file corresponding to a particular restart file index, which can then be passed to `PHiLiP` (in place of the original `.prm` file), _without any manual modifications required_, to restart the simulation. Necessary documentation provided in `INSTALL.md`.

Tests added: 
- `MPI_VISCOUS_TAYLOR_GREEN_VORTEX_RESTART_CHECK`
- `MPI_VISCOUS_TAYLOR_GREEN_VORTEX_RESTART_FROM_PARAMETER_FILE_CHECK`

**Note:** This restart capability is implemented and tested for unsteady flow cases; the extension to steady state flows could be developed in the future if needed. Enhancement issue opened for this. 

Additional contributions:
- Minor cleanup in `ODESolverBase`, and `FiniteDifferenceSensitivity`
- Updating `get_NACA0012_mesh_files_cluster.sh`

**DISCLAIMER: Approximately 1.4k changes are solely from the addition of a `.prm` file that includes all default parameters.** This is necessary for the added test: `MPI_VISCOUS_TAYLOR_GREEN_VORTEX_RESTART_FROM_PARAMETER_FILE_CHECK`


src/testing/flow_solver.cpp
```diff
@@ -1,35 +1,324 @@
...
     solution_no_ghost.reinit(dg->locally_owned_dofs, MPI_COMM_WORLD);
-    dealii::VectorTools::interpolate(dg->dof_handler, *initial_condition_function, solution_no_ghost);
-    dg->solution = solution_no_ghost;
+
+    if(flow_solver_param.restart_computation_from_file == true) {
+        if(dim == 1) {
+            pcout << "Error: restart_computation_from_file is not possible for 1D. Set to false." << std::endl;
+            std::abort();
+        }
+#if PHILIP_DIM>1
```
dougshidong, line 45: I'd guess this has to do with SolutionTransfer not existing for 1D?

Move it such that only the part that is dim-dependent is in your compiler #if .

src/testing/flow_solver_cases/1D_burgers_rewienski_snapshot.cpp
```diff
@@ -118,7 +118,7 @@ void BurgersRewienskiSnapshot<dim, nstate>::steady_state_postprocessing(std::sha
                 sensitivity_dWdb[i]);
     }
     solutions_table.set_precision("Sensitivity:", 16);
-    std::ofstream out_file(this->all_param.flow_solver_param.sensitivity_table_filename + ".txt");
+    std::ofstream out_file(this->unsteady_data_table_filename_with_extension);
```
donovan97, line 121: This should be left to the value it was before, sensitivity_table_filename.

src/testing/flow_solver_cases/flow_solver_case_base.cpp
```diff
@@ -7,10 +7,11 @@ template<int dim, int nstate>
 FlowSolverCaseBase<dim, nstate>::FlowSolverCaseBase(const PHiLiP::Parameters::AllParameters *const parameters_input)
         : all_param(*parameters_input)
         , mpi_communicator(MPI_COMM_WORLD)
-        , mpi_rank (dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD))
-        , n_mpi (dealii::Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD))
+        , mpi_rank(dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD))
+        , n_mpi(dealii::Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD))
         , pcout(std::cout, mpi_rank==0)
-          {}
+        , unsteady_data_table_filename_with_extension(all_param.flow_solver_param.unsteady_data_table_filename+".txt")
```
donovan97, line 13: Why move this to flow solver case base? Not all flow solver cases will necessarily require it.

src/testing/flow_solver.cpp
```diff
@@ -1,35 +1,324 @@
...
+        dealii::parallel::distributed::SolutionTransfer<dim, dealii::LinearAlgebra::distributed::Vector<double>, dealii::DoFHandler<dim>> solution_transfer(dg->dof_handler);
+        solution_transfer.deserialize(solution_no_ghost);
+#endif
+    } else {
+        // Initialize solution from initial_condition_function
+        pcout << "Initializing solution with initial condition function... " << std::flush;
+        dealii::VectorTools::interpolate(dg->dof_handler, *initial_condition_function, solution_no_ghost);
+    }
+    dg->solution = solution_no_ghost; //< assignment
+    // dg->solution.update_ghost_values(); // -- should this be done ?
```
dougshidong, line 67: Maybe. Safer to do it and inexpensive. DG should automatically update the ghost values if you evaluate the residual. However, if you do something else with the solution that requires ghost values, then this is needed.

src/testing/flow_solver.cpp
```diff
@@ -1,35 +1,324 @@
...
+         */
+        std::vector<std::string> subsection_line;
+        subsection_line.push_back("subsection ODE solver");
+        subsection_line.push_back("subsection flow_solver");
+        // Number of subsections to change values in
+        int number_of_subsections = subsection_line.size();
+
+
+        /* WARNING: (1) Must put a space before and after each parameter string as done below
+         *          (2) These must be in the order they appear in the .prm file
```
dougshidong, line 197: What happens if either of those warnings aren't satisfied. Is there a way to detect this and tell the user without the user knowing about this comment in the code?

src/testing/flow_solver.cpp
```diff
@@ -1,35 +1,324 @@
...
+    const double constant_time_step,
+    const std::shared_ptr <dealii::TableHandler> unsteady_data_table) const
+{
+    pcout << "  ... Writing restart files ... " << std::endl;
+    const std::string restart_filename_without_extension = get_restart_filename_without_extension(current_restart_index);
+
+    // solution files
+    dealii::parallel::distributed::SolutionTransfer<dim, dealii::LinearAlgebra::distributed::Vector<double>, dealii::DoFHandler<dim>> solution_transfer(dg->dof_handler);
+    solution_transfer.prepare_for_coarsening_and_refinement(dg->solution);
+    dg->triangulation->save(flow_solver_param.restart_files_directory_name + std::string("/") + restart_filename_without_extension);
```
dougshidong, line 308: I'm looking at
https://www.dealii.org/developer/doxygen/deal.II/classparallel_1_1distributed_1_1SolutionTransfer.html
subsection Use for Serialization

Is prepare_for_serialization() needed?

src/testing/flow_solver.cpp
dougshidong: As discussed in chat, those warnings are for future changes. A user of the restart capability won't encounter any issues. No need to fail/warn since someone who changes the code will necessarily read the above warning.

dougshidong (approved):
LGTM after the comments have been addressed. Can you open up an issue regarding the possible separation of .prm files as we had discussed. And another issue for @PranshulThakur to look into for the huge chunk of lines in the .prm due to the grid refinement subsection.

dougshidong:
Also added @PranshulThakur, in case he spots something about future changes with hp-refinement.

src/testing/flow_solver_cases/flow_solver_case_base.cpp
jbrillon: Maybe we should just replace `sensitivity_table_filename` and `unsteady_data_table_file_name` by `data_table_filename`? This way we don't have case specific variables in the flow solver case base class

src/testing/flow_solver_cases/1D_burgers_rewienski_snapshot.cpp
jbrillon: My mistake, sorry about that

src/testing/flow_solver.cpp
jbrillon: I think so, I've changed it and the tests are still passing. `prepare_for_coarsening_and_refinement` was used in other parts of the code so that's why I had used it in the first place. Take a look at subsection "Note on usage with [DoFHandler](https://www.dealii.org/current/doxygen/deal.II/classDoFHandler.html) with hp-capabilities". It looks like that's what were actually doing since we're passing in that 3rd template variable; I'm passing it because it was done like that in other parts of the code; see line 30 of `src/grid_refinement/grid_refinement_uniform.cpp`

src/testing/flow_solver_cases/flow_solver_case_base.cpp
donovan97: I would agree except that I need both, since one of my tests is outputting the solution and a sensitivity. Thats why I needed to create a second filename parameter.

src/testing/flow_solver_cases/1D_burgers_rewienski_snapshot.cpp
jbrillon: I meant to make this change in `compute_unsteady_data_and_write_to_table` within this file; I've address this

src/testing/flow_solver_cases/flow_solver_case_base.cpp
jbrillon: Actually, you do use it in `compute_unsteady_data_and_write_to_table` in both your `BurgersRewienskiSnapshot` and `BurgersViscousSnapshot` classes so I think we should leave it -- but can consider renaming the variable it if you want

src/testing/flow_solver.cpp
jbrillon: addressed in most recent commit

src/testing/flow_solver.cpp
jbrillon: Thanks for the explanation, I've addressed this in the most recent commit

src/testing/flow_solver.cpp
dougshidong: Funnily enough the deal.II internally calls prepare_for_coarsening_and_refinement, however it packages it in a std::vector of vector in case you want to serialize multiple solutions.
https://github.com/dougshidong/dealii/blob/6abe6d7889e33a788aa30b29d98e7c6b205618c2/source/distributed/solution_transfer.cc#L180-L197

However, I think it makes sense to use prepare_for_serialization() instead because the function name is more relevant.

src/testing/flow_solver_cases/flow_solver_case_base.cpp
donovan97: I was more looking at it from the point of view that not every future test will necessarily be computing unsteady data and writing to a table so it probably doesn't make sense to have a such a specific variable in the flow solver case base class. Renaming it to something more generic would probably make more sense

jbrillon: I've addressed this in the most recent commit, I simply removed it from `FlowSolverCaseBase` and put it in `PeriodicCubeFlow`; and restored your Burgers classes how they originally were

donovan97 (approved):

dougshidong:
Looks good to go. Anything else you wanted to add @jbrillon?

jbrillon:
@dougshidong No more additions, go ahead and merge it

PranshulThakur (approved):
Looks good to me. I will open a separate issue for the prm files from grid refinement.

Issue #121, opened 2022-04-13
Uncomment test 2D_NAVIER_STOKES_INTEGRATION_NACA0012_SUBSONIC_LONG


jbrillon:
Uncomment integration test `2D_NAVIER_STOKES_INTEGRATION_NACA0012_SUBSONIC_LONG` added in PR #116 once @sivanadarajah has merged his modifications to the linear solver that ended up making this test reproduce the Swanson NACA0012 test case with the recirculation region. 

Reference to Swanson test case:
@inproceedings{swanson1985multistage,
  title={A multistage time-stepping scheme for the Navier-Stokes equations},
  author={Swanson, R and Turkel, Eli},
  booktitle={23rd Aerospace Sciences Meeting},
  pages={35},
  year={1985}
}

PR #134, opened 2022-04-27
Naca0012 flowsolver

donovan97:
Pull request to add NACA0012 case to `FlowSolver`. 

Main change outside of FlowSolver and related classes is in initial_condition.cpp, to replace nstate variable by a template parameter so that the initial conditions factory is compatible with FreeStreamInitialConditions.

I have also reverted the max refinements to 10, as all the Burger Rewienski tests require this and would currently be failing in the master branch.


src/physics/initial_conditions/initial_condition.h
```diff
@@ -32,38 +48,22 @@ class FreeStreamInitialConditions : public dealii::Function<dim>
         for (int d=0;d<dim;d++) { primitive_boundary_values[1+d] = euler_physics.velocities_inf[d]; }
         primitive_boundary_values[nstate-1] = pressure_bc;
         farfield_conservative = euler_physics.convert_primitive_to_conservative(primitive_boundary_values);
+        std::cout << "Farfield conditions: "<< std::endl;
+        for (int s=0;s<nstate;s++) {
+            std::cout << farfield_conservative[s] << std::endl;
+        }
```
jbrillon, line 54: When running with mpi, does this output np times? If so, we'd want to avoid that and consider printing it directly in the FlowSolverCase derived class which has knowledge of `mpi_rank`

src/physics/initial_conditions/initial_condition.cpp
```diff
@@ -169,15 +172,23 @@ InitialConditionFactory<dim,real>::create_InitialConditionFunction(
 // ========================================================
 // ZERO INITIAL CONDITION
 // ========================================================
-template<int dim, typename real>
-real InitialConditionFunction_Zero<dim, real> :: value(const dealii::Point<dim,real> &/*point*/, const unsigned int /*istate*/) const 
+template <int dim, int nstate, typename real>
+real InitialConditionFunction_Zero<dim, nstate, real> :: value(const dealii::Point<dim,real> &/*point*/, const unsigned int /*istate*/) const
```
jbrillon, line 176: Just wondering, how come `InitialConditionFunction_Zero` is not in the `create_InitialConditionFunction()` ?

src/testing/euler_naca0012.cpp
```diff
@@ -206,7 +204,7 @@ ::run_test () const
             //std::shared_ptr<HighOrderGrid<dim,double>> joukowski_mesh = read_gmsh <dim, dim> ("joukowski_R1_Q"+std::to_string(grid_degree)+".msh");
             //std::shared_ptr<HighOrderGrid<dim,double>> naca0012_mesh = read_gmsh <dim, dim> ("new_msh41.msh");
             //std::shared_ptr<HighOrderGrid<dim,double>> naca0012_mesh = read_gmsh <dim, dim> ("naca0012_hopw.msh");
-            std::shared_ptr<HighOrderGrid<dim,double>> naca0012_mesh = read_gmsh <dim, dim> (mesh_filename);
+            std::shared_ptr<HighOrderGrid<dim,double>> naca0012_mesh = read_gmsh <dim, dim> ("naca0012.msh");
```
jbrillon, line 207: Revert this change

jbrillon: Different meshes can be used, like the test that is to be uncommented in issue #121

jbrillon: Unless the plan is to move that test to FlowSolver

tests/integration_tests_control_files/euler_integration/naca0012/CMakeLists.txt
```diff
@@ -12,6 +12,13 @@ add_test(
   WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
 )
 
+configure_file(naca0012.prm naca0012.prm COPYONLY)
+add_test(
+        NAME NACA0012_FLOWSOLVER
```
jbrillon, line 17: Maybe give this a better name that's a bit more descriptive, like `INVISCID_TRANSONIC_STEADY_STATE_NACA0012`

tests/integration_tests_control_files/euler_integration/naca0012/naca0012.prm
```diff
@@ -0,0 +1,70 @@
+# Listing of Parameters
+# ---------------------
+
```
jbrillon, line 3: Following my comment regarding the test name for this, also rename this `.prm` file appropriately

src/testing/flow_solver_cases/naca0012.cpp
```diff
@@ -0,0 +1,116 @@
...
+    grid->refine_global();
+
+    return grid;
+}
+
+template <int dim, int nstate>
+void NACA0012<dim,nstate>::set_higher_order_grid(std::shared_ptr<DGBase<dim, double>> dg) const
+{
+    const std::string mesh_filename = this->all_param.flow_solver_param.input_mesh_filename+std::string(".msh");
+    std::shared_ptr<HighOrderGrid<dim,double>> naca0012_mesh = read_gmsh<dim, dim> (mesh_filename);
```
jbrillon, line 84: I see that it's done here now, I'll resolve the other comment

jbrillon (commented):
Looks great, just a few comments to address. Very useful PR, this is the first aerodynamic case added to the FlowSolver. General comment, since this is now packaged within the FlowSolver class, we could significantly simplify existing tests like `src/testing/euler_naca0012.cpp` etc by doing something like `flow_solver = create_FlowSolver` inside then doing `flow_solver->run_test()`; similar to how its done in `src/testing/taylor_green_vortex_energy_check.cpp`; consider opening an issue for this, listing all the relevant `src/testing/` files that can be simplified

dougshidong:
Added @cpethrick since she is currently getting up to speed to FlowSolver.

Also will be adding @MayaTatarelli after the invitation to collab has been accepted since she might be running more aerodynamic cases this summer.

Even though they might both not be familiar with the current structure, it's a good opportunity to get exposure on his side of the code.

src/testing/flow_solver_cases/flow_solver_case_base.h
```diff
@@ -34,6 +34,9 @@ class FlowSolverCaseBase
     /// Pure Virtual function to generate the grid
     virtual std::shared_ptr<Triangulation> generate_grid() const = 0;
 
+    /// Set higher order grid
```
dougshidong, line 37: Seems like the default implementation is an empty function? Did you mean to set is a pure virtual function (=0)? Then it seems like the only derived class that uses that function is the naca0012 case.

Should this only be defined in the naca0012 case?

src/testing/euler_naca0012.cpp
donovan97: It up to you if you want to move the test to FlowSolver and get rid of euler_naca0012.cpp. The only thing is that there won't be any more grid study or polynomial ramping, since FlowSolver isn't really set up for that.

src/testing/flow_solver_cases/naca0012.cpp
```diff
@@ -0,0 +1,116 @@
...
+    airfoil_data.bias_factor    = 5.0;
+    airfoil_data.refinements    = 0;
+    airfoil_data.n_subdivision_x_0 = 4;
+    airfoil_data.n_subdivision_x_1 = 4;
+    airfoil_data.n_subdivision_x_2 = 4;
+    airfoil_data.n_subdivision_y = 4;
+
+    airfoil_data.airfoil_sampling_factor = 100000; // default 2
+
+    Grids::naca_airfoil(reinterpret_cast<dealii::parallel::distributed::Triangulation<2> &>(*grid), airfoil_data);
```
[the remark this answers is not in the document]
donovan97: Not entirely sure, there might be something wrong with the templates. The error I'm getting if I don't use the reinterpret_cast is: 

error: invalid initialization of reference of type ‘dealii::parallel::distributed::Triangulation<2>&’ from expression of type ‘std::__shared_ptr_access<dealii::parallel::distributed::Triangulation<3>, __gnu_cxx::_S_atomic, false, false>::element_type’ {aka ‘dealii::parallel::distributed::Triangulation<3>’}

Maybe you have a better idea of why that's happening and how to prevent it?

src/physics/initial_conditions/initial_condition.h
donovan97: It does output np times. Unfortunately, the FlowSolverCase derived classes have no knowledge of the initial conditions. 

How important do you think it is to output that information? If it's important enough, I'll just add the mpi communicator to the InitialConditionFunction class and use pcout instead.

src/testing/flow_solver_cases/flow_solver_case_base.h
donovan97: If it were a pure virtual function, all flow solver cases would have to implement it, which is not necessarily the case as not all cases have a higher order grid. 

The other tests currently only use the generate_grid function, which is pure virtual. The problem with using that function in this case is that adding the higher order grid requires having access to dg, which generate_grid does not. So I can either modify generate_grid to take a dg (and then comment out /*dg*/ in all tests that don't require a dg to set the grid). Or I can make a new function that always gets called by FlowSolver, and is set to do nothing in FlowSolverCaseBase, but if needed can be overrided with the appropriate behaviour.

donovan97:
> Looks great, just a few comments to address. Very useful PR, this is the first aerodynamic case added to the FlowSolver. General comment, since this is now packaged within the FlowSolver class, we could significantly simplify existing tests like `src/testing/euler_naca0012.cpp` etc by doing something like `flow_solver = create_FlowSolver` inside then doing `flow_solver->run_test()`; similar to how its done in `src/testing/taylor_green_vortex_energy_check.cpp` and in `src/testing/taylor_green_vortex_restart_check.cpp` with the parameter objects; consider opening an issue for this, listing all the relevant `src/testing/` files that can be simplified

Sure, I'll open an issue once we merge to master.

src/testing/flow_solver_cases/naca0012.h
```diff
@@ -0,0 +1,51 @@
...
+#include "flow_solver_case_base.h"
+
+namespace PHiLiP {
+namespace Tests {
+
+#if PHILIP_DIM==1
+    using Triangulation = dealii::Triangulation<PHILIP_DIM>;
+#else
+    using Triangulation = dealii::parallel::distributed::Triangulation<PHILIP_DIM>;
+#endif
```
dougshidong, line 21: Move those using inside the class. Otherwise those using become "global" to the PHiLiP::Tests namespace when this file is included.

tests/integration_tests_control_files/euler_integration/naca0012/naca0012.prm
donovan97: Fixed in most recent commit.

tests/integration_tests_control_files/euler_integration/naca0012/CMakeLists.txt
donovan97: Fixed in most recent commit.

src/testing/flow_solver_cases/naca0012.cpp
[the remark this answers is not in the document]
dougshidong: Seems like it thinks `grid` is a 3D Triangulation, but the function only works for 2D Triangulations. The reinterpret cast makes the 3D Triangulation being read as a 2D Triangulation, which is no good either.

I'm suprised it generates this error since you seem to only instantiate this class for PHILIP_DIM==2. Is the comment about moving the "using" helping fix this?

src/testing/flow_solver_cases/naca0012.h
donovan97: Fixed in most recent commit.

src/testing/flow_solver_cases/naca0012.cpp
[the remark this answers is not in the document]
donovan97: No, moving the usings did not help.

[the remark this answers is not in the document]
dougshidong: Can you try `make 2D` and or `make 3D`, and see if the error pops up in both of them? I would expect it to only fail compilation for `make 3D`.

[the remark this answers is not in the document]
donovan97: It fails only for make 3D. It also fails for make 1D, with a similar error message.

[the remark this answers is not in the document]
dougshidong: How about a:
if constexpr(dim==2) {
    Grids::naca_airfoil()
} else {
    // Error message
}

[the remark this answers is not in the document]
dougshidong: This is an equivalent of compiler #if PHILIP_DIM==2, but prettier. I think that the ideal solution is to allow Grids::naca_airfoil() to take in a generic grid dimension (and fail appropriately). But this solution is fine since we'll never run that flow case for 3D.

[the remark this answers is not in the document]
donovan97: Using the constexpr does not work and gives the same error message, since it's at compile time. Using the #if preprocessor directive does work. So I guess the solution is:

#if PHILIP_DIM==2
Grids::naca_airfoil()
#endif

If you're good with that, I'll commit it.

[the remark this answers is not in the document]
dougshidong: Sure. Better than than the reinterpret cast. The reinterpret_cast is pretty "evil" and usually done for low-level programming C-style.

Kinda annoyed that it gets a compilation error for 1 and 3 dimension when it shouldn't even be instantiated/compiled.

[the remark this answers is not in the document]
donovan97: Following Discord conversation, I have replaced the entire function with something simpler, since it's just a dummy triangulation. This also solves the issue here.

dougshidong:
Looks good to me. Comments I left are prime examples of issues arising from my past self leaving dead code which causes further confusion.

src/physics/initial_conditions/initial_condition.cpp
jbrillon: Ok I think it's fine then

src/physics/initial_conditions/initial_condition.h
jbrillon: I think what would be a good solution would be to pass `initial_condition_function` as an argument to `flow_solver_case->display_flow_solver_setup();` in the constructor of `flow_solver.cpp`, then loop over the `nstate` in your definition of `display_flow_solver_setup` for the naca0012 case to display it. This will be useful for future cases we add as well and sounds like the most appropriate place to be printing this stuff. What do you think? Just make it an unused argument in the other FlowSolverCaseBase derived classes

src/testing/euler_naca0012.cpp
jbrillon: What I meant was, replace the section of code inside `euler_naca0012.cpp` that essentially executes what the flow solver `run_test()` function does. This way the setup and time advancement is all accomplished inside there. Is this possible?  Like in the for loop you create or update an `all_parameters` object, then do `create_FlowSolver(all_parameters)` then `flow_solver->run_test()` and use the updated members like `flow_solver->dg` to do the rest of the operations in the `euler_naca0012.cpp` test

tests/integration_tests_control_files/euler_integration/naca0012/CMakeLists.txt
```diff
@@ -12,6 +12,13 @@ add_test(
   WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
 )
 
+configure_file(inviscid_transonic_steady_state_naca0012.prm inviscid_transonic_steady_state_naca0012.prm COPYONLY)
+add_test(
+        NAME INVISCID_TRANSONIC_STEADY_STATE_NACA0012
+        COMMAND mpirun -np ${MPIMAX} ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_2D -i ${CMAKE_CURRENT_BINARY_DIR}/naca0012.prm
```
jbrillon, line 18: I think the argument should be `inviscid_transonic_steady_state_naca0012.prm` here

src/testing/flow_solver_cases/naca0012.cpp
```diff
@@ -0,0 +1,95 @@
...
+    using PDE_enum = Parameters::AllParameters::PartialDifferentialEquation;
+    const PDE_enum pde_type = this->all_param.pde_type;
+    const double pi = atan(1.0) * 4.0;
+    std::string pde_string;
+    if (pde_type == PDE_enum::euler)                {pde_string = "euler";}
+    if (pde_type == PDE_enum::navier_stokes)        {pde_string = "navier_stokes";}
+    this->pcout << "- PDE Type: " << pde_string << std::endl;
+    this->pcout << "- Polynomial degree: " << this->all_param.grid_refinement_study_param.poly_degree << std::endl;
+    this->pcout << "- Courant-Friedrich-Lewy number: " << this->all_param.flow_solver_param.courant_friedrich_lewy_number << std::endl;
+    this->pcout << "- Freestream Reynolds number: " << this->all_param.navier_stokes_param.reynolds_number_inf << std::endl;
```
jbrillon, line 42: This should only be displayed if `pde_type == PDE_enum::navier_stokes`

src/testing/flow_solver_cases/naca0012.cpp
```diff
@@ -0,0 +1,95 @@
...
+    this->pcout << "- Freestream Mach number: " << this->all_param.euler_param.mach_inf << std::endl;
+    this->pcout << "- Angle of attack: " << this->all_param.euler_param.angle_of_attack*180/pi << std::endl;
+    this->pcout << "- Side-slip angle: " << this->all_param.euler_param.side_slip_angle*180/pi << std::endl;
+}
+
+template <int dim, int nstate>
+std::shared_ptr<Triangulation> NACA0012<dim,nstate>::generate_grid() const
+{
+    //Dummy triangulation
+    std::shared_ptr<Triangulation> grid = std::make_shared<Triangulation>(MPI_COMM_WORLD);
```
jbrillon, line 52:
```suggestion
    std::shared_ptr<Triangulation> grid = std::make_shared<Triangulation>(this->mpi_communicator);
```

src/testing/flow_solver_cases/naca0012.cpp
```diff
@@ -0,0 +1,95 @@
...
+    dealii::GridGenerator::Airfoil::AdditionalData airfoil_data;
+    dealii::GridGenerator::Airfoil::create_triangulation(*grid, airfoil_data);
+    grid->refine_global();
+    return grid;
```
jbrillon, line 56: Also, can this be moved to `src/mesh/grids/` and then you call it here instead? We want the `FlowSolverCaseBase` classes to simply call the appropriate grid from the files in that directory; see `src/flow_solver_cases/periodic_cube_flow.cpp`. Ideally we'd have a grid factory where we just pass the parameters and it returns the grid, that way we'd have only one `generate_grid` definition that would be in the base class. There seems to be a file `src/mesh/grids/naca_airfoil_grip.cpp`, can we use that? @dougshidong

src/testing/flow_solver_cases/naca0012.cpp
```diff
@@ -0,0 +1,95 @@
...
+    dg->set_high_order_grid(naca0012_mesh);
+    for (unsigned int i=0; i<this->all_param.grid_refinement_study_param.num_refinements; ++i) {
+        dg->high_order_grid->refine_global();
+    }
+    const unsigned int n_global_active_cells = dg->triangulation->n_global_active_cells();
+    const unsigned int n_dofs = dg->dof_handler.n_dofs();
+    this->pcout << "Dimension: " << dim << "\t Polynomial degree p: " << this->all_param.grid_refinement_study_param.poly_degree << std::endl
+          << ". Number of active cells: " << n_global_active_cells
+          << ". Number of degrees of freedom: " << n_dofs
+          << std::endl;
```
cpethrick, line 84: It's a bit inconsistent that you're defining variables in some places (`n_dofs` and `n_global_active_cells`) but not in other places (`all_param.grid_refinement_study_param.num_refinements`, for instance).

src/testing/flow_solver_cases/naca0012.cpp
donovan97: This grid isn't actually used, its just a dummy triangulation because dg needs it to be initialized. Doug opened issue #135 for this. I was also having some issues yesterday with the naca_airfoil function in naca_airfoil_grid.cpp, you can see one of the conversations with Doug about it. We probably shouldn't bother since its just a dummy triangulation anyway.

src/physics/initial_conditions/initial_condition.h
donovan97: Sure, done is most recent commit.

tests/integration_tests_control_files/euler_integration/naca0012/inviscid_transonic_steady_state_naca0012.prm
```diff
@@ -0,0 +1,70 @@
...
+  set initial_time_step = 1e3
+  set time_step_factor_residual = 15.0
+  set time_step_factor_residual_exp = 2
+  set print_iteration_modulo              = 1
+end
+
+subsection grid refinement study
+ set poly_degree = 0
+ set num_refinements = 0
+ #set grid_size = 4
```
cpethrick, line 62: Remove commented parameter here?

tests/integration_tests_control_files/euler_integration/naca0012/CMakeLists.txt
donovan97: Good catch, fixed in most recent commit.

src/testing/flow_solver_cases/naca0012.cpp
donovan97: Fixed in most recent commit.

tests/integration_tests_control_files/flow_solver/CMakeLists.txt
```diff
@@ -27,10 +27,10 @@ set(TEST_OUTPUT_DIR ${CMAKE_CURRENT_BINARY_DIR})
...
-#  NAME MPI_VISCOUS_TAYLOR_GREEN_VORTEX
-#  COMMAND mpirun -np ${MPIMAX} ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_3D -i ${CMAKE_CURRENT_BINARY_DIR}/viscous_taylor_green_vortex.prm
-#  WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
-#)
+configure_file(viscous_taylor_green_vortex.prm viscous_taylor_green_vortex.prm COPYONLY)
+add_test(
+  NAME MPI_VISCOUS_TAYLOR_GREEN_VORTEX
+  COMMAND mpirun -np ${MPIMAX} ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_3D -i ${CMAKE_CURRENT_BINARY_DIR}/viscous_taylor_green_vortex.prm
+  WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
+)
```
cpethrick, line 35: Your PR doesn't say anything about adding this test. Was this file possibly committed by accident? If it's intentional, maybe add to the description of your PR.

tests/integration_tests_control_files/reduced_order/CMakeLists.txt
```diff
@@ -46,12 +46,12 @@ add_test(
...
-#        NAME 1D_BURGERS_REWIENSKI_SENSITIVITY
-#        COMMAND mpirun -np 1 ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_1D -i ${CMAKE_CURRENT_BINARY_DIR}/1d_burgers_rewienski_fd_sensitivity.prm
-#        WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
-#)
+configure_file(1d_burgers_rewienski_fd_sensitivity.prm 1d_burgers_rewienski_fd_sensitivity.prm COPYONLY)
+add_test(
+        NAME 1D_BURGERS_REWIENSKI_SENSITIVITY
+        COMMAND mpirun -np 1 ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_1D -i ${CMAKE_CURRENT_BINARY_DIR}/1d_burgers_rewienski_fd_sensitivity.prm
+        WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
+)
```
cpethrick, line 54: Likewise here, this test has been uncommented.

cpethrick (commented):
Good PR! I think the change of templating nstate is a good addition to the flow solver, your changes make things much cleaner. I left a few comments, including some tests which were previously commented - I'm unsure whether this is an intentional change.

src/testing/flow_solver_cases/naca0012.cpp
```diff
@@ -0,0 +1,106 @@
...
+        this->pcout << "- Freestream Reynolds number: " << this->all_param.navier_stokes_param.reynolds_number_inf << std::endl;
+    }
+    this->pcout << "- Courant-Friedrich-Lewy number: " << this->all_param.flow_solver_param.courant_friedrich_lewy_number << std::endl;
+    this->pcout << "- Freestream Mach number: " << this->all_param.euler_param.mach_inf << std::endl;
+    this->pcout << "- Angle of attack: " << this->all_param.euler_param.angle_of_attack*180/pi << std::endl;
+    this->pcout << "- Side-slip angle: " << this->all_param.euler_param.side_slip_angle*180/pi << std::endl;
+    this->pcout << "- Farfield conditions: " << std::endl;
+    const dealii::Point<dim> dummy_point;
+    for (int s=0;s<nstate;s++) {
+        this->pcout << initial_condition->value(dummy_point, s) << std::endl;
```
jbrillon, line 51:
```suggestion
        this->pcout << "  - State " << s << "; Value: " << initial_condition->value(dummy_point, s) << std::endl;
```

src/testing/euler_naca0012.cpp
donovan97: Check out my most recent commit.

src/testing/flow_solver_cases/naca0012.cpp
donovan97: Fixed in recent commit.

tests/integration_tests_control_files/euler_integration/naca0012/inviscid_transonic_steady_state_naca0012.prm
donovan97: Fixed in recent commit.

tests/integration_tests_control_files/flow_solver/CMakeLists.txt
donovan97: I was just making sure that the test was working properly after changing the initial conditions class. But thanks, I have re-commented it out.

tests/integration_tests_control_files/reduced_order/CMakeLists.txt
donovan97: Also commented out in recent commit.

tests/integration_tests_control_files/flow_solver/CMakeLists.txt
cpethrick: Okay, looks good!

dougshidong:
I think we've got enough eyes on this small PR.

Reviewers can resolve their comment when satisfied. @MayaTatarelli , feel free to press "Squash and merge" whenever you're happy with it.

jbrillon (approved):
Great work on simplifying `euler_naca0012.cpp`!  It's great to see the `FlowSolver` class being utilized as its supposed to, more tests can be simplified in the future.

MayaTatarelli (approved):
Great work @donovan97 !

PR #132, opened 2022-04-26
Grid Refinement Parameters Fix

jbrillon:
Pull request to close issue #130 

`MAX_REFINEMENTS` was initially set to 10, while it was found that only 4 was required after investigating the code. This saves 420 lines in our parameter files.

PR #136, opened 2022-04-29
Correcting MAX_REFINEMENTS

jbrillon:
PR to correct #132; changing MAX_REFINEMENTS back to 10 as instructed by @donovan97 ; some of his tests will fail without it set to 10.


donovan97 (approved):
Great work!

jbrillon:
Done in PR #134

Issue #115, opened 2022-03-28
Time Refinement Study and Related Code Restructuring


cpethrick:
This issue will aim to create a general time refinement study for explicit RK timestepping.
- Restructure the time refinement study located at `tests/unit_tests/ode_solver_unit_test/explicit_ode_solver.cpp` to use the `flow_solver` framework and move to `src/testing` with other integration tests
- Move grid generation of aforementioned test to `straight_periodic_cube` such that 1D, 2D and 3D periodic cube (or square or line) domains are generated by the same code
- Move initial condition into `src/physics/initial_conditions`
- Create a framework for generating exact solutions in the same way that initial conditions are generated

PR #138, opened 2022-05-05
Time Refinement Study using flow_solver

cpethrick:
Closes issue #115. Changes are as follows:

- Added a new test, `1D_TIME_REFINEMENT_STUDY_ADVECTION_EXPLICIT`, which uses the `flow_solver.cpp` framework. This test replaces `1D_explicit_ode_solver' test (which has been removed by removing the directory from `CMakeLists.txt` in 'tests/unit_tests`). Results are the same.
- Domain setup is in `periodic_1D_unsteady.cpp`
- Mesh generation in `straight_periodic_cube.cpp` has been extended from only 3D to any dimension
- Defined exact solutions in `physics/exact_solutions`, similar to how initial conditions are defined. The function `create_ExactSolutionFunction` selects an exact solution based on the flow solver type, and defines it at the `final_time` which is passed.


src/testing/flow_solver_cases/periodic_1D_flow.cpp
```diff
@@ -0,0 +1,125 @@
...
+template <int dim, int nstate>
+std::shared_ptr<Triangulation> Periodic1DFlow<dim,nstate>::generate_grid() const
+{
+    std::shared_ptr<Triangulation> grid = std::make_shared<Triangulation> (
+#if PHILIP_DIM!=1
+            this->mpi_communicator
+#endif
+    );
+    
+    dealii::GridGenerator::hyper_cube(*grid, domain_left, domain_right, true); //colorize = true
```
jbrillon, line 60: This could be moved to `src/mesh/grids/straight_periodic_cube.cpp`

src/testing/flow_solver_cases/periodic_1D_flow.cpp
```diff
@@ -0,0 +1,125 @@
...
+    this->pcout << "- - Domain dimensionality: " << dim << std::endl;
+    this->pcout << "- - Domain left: " << domain_left << std::endl;
+    this->pcout << "- - Domain right: " << domain_right << std::endl;
+    this->pcout << "- - Number of refinements: " << number_of_refinements << std::endl;
+
+    return grid;
+}
+
+template <int dim, int nstate>
+double Periodic1DFlow<dim,nstate>::get_constant_time_step(std::shared_ptr<DGBase<dim,double>> /* dg*/) const
```
jbrillon, line 78: Can remove this function from the class since it is the same as the definition in the base class `flow_solver_case_base.cpp`

src/testing/flow_solver_cases/periodic_1D_flow.cpp
jbrillon: Consider modifying the existing `periodic_cube_flow.cpp` and using the `is_taylor_green_vortex` flag to turn off things that do now pertain to your flow case because there is a lot of repeated code

src/testing/time_refinement_study_advection.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+template <int dim, int nstate>
+ Parameters::AllParameters TimeRefinementStudyAdvection<dim, nstate>::reinit_params_and_refine_timestep(int refinement) const{
+     dealii::ParameterHandler parameter_handler;
+     PHiLiP::Parameters::AllParameters::declare_parameters (parameter_handler);
+     PHiLiP::Parameters::AllParameters parameters;
+     parameters.parse_parameters(parameter_handler);
+
+
+     //parameters = this->all_parameters; //This doesn't work, operator = not defined. Block below copies params one at a time.
```
jbrillon, line 27: See how it's done in `src/testing/taylor_green_vortex_restart_check.cpp` function `reinit_params()`

dougshidong:
Wondered if there was a need for `physics/exact_solutions` as per previous discussion on Alex's PR about adding a time component to either manufactured solutions or initial conditions. Also, manufactured solutions and initial conditions are already very similar in that they basically just override the dealii::Function which already has a way to modify a time component 

https://www.dealii.org/developer/doxygen/deal.II/classFunction.html
https://www.dealii.org/developer/doxygen/deal.II/classFunctionTime.html#a3a583fd8f30db3549dbaa43b2592e2bc

At this point, I'm wondering if we even need to have different classes such as ManufacturedSolution, InitialConditions, and ExactSolution, and if we should instead just have a single SolutionFunctionFactory for all three (especially since some initial conditions are just the exact solutions at time t = 0).

src/testing/flow_solver.cpp
```diff
@@ -323,6 +326,41 @@ void FlowSolver<dim,nstate>::output_restart_files(
 }
 #endif
 
+template <int dim, int nstate>
+double FlowSolver<dim,nstate>::calculate_L2_error_at_final_time_wrt_function()
```
jbrillon, line 330: I think it makes more sense for this to be inside the `time_refinement_study` TestBase class and simply pass a `dg` object and `final_time`, which you can both get from the `flow_solver` object in that class, after doing `flow_solver->run_test()`.

src/testing/time_refinement_study_advection.cpp
```diff
@@ -0,0 +1,135 @@
+#include "time_refinement_study_advection.h"
+#include "flow_solver.h"
+#include "flow_solver_cases/periodic_1D_flow.h"
+#include "cmath"
+
+namespace PHiLiP {
+namespace Tests {
+
+template <int dim, int nstate>
+TimeRefinementStudyAdvection<dim, nstate>::TimeRefinementStudyAdvection(
```
jbrillon, line 10: This class looks quite general in terms of handling any PDE type, if thats the case, it could be renamed `TimeRefinementStudy` or `TemporalConvergenceStudy`?

src/testing/time_refinement_study_advection.cpp
```diff
@@ -0,0 +1,135 @@
...
+     
+     parameters.ode_solver_param.initial_time_step *= pow(refine_ratio,refinement);
+     //ADD : flow_solver params: change file name
+     return parameters;
+ }
+
+
+/*
+template <int dim, int nstate>
+void TimeRefinementStudyAdvection<dim, nstate>::write_convergence_summary(int refinement, double dt) {
```
jbrillon, line 57: Should this be commented? Remove if it won't get used

src/testing/time_refinement_study_advection.h
```diff
@@ -0,0 +1,47 @@
...
+    
+    /// Run test
+    int run_test () const override;
+protected:
+
+    /// Number of times to solve for convergence summary
+    const int n_time_calculations = 4;
+
+    /// Ratio to refine by
+    const double refine_ratio = 0.5;
```
jbrillon, line 34: Consider making these input parameters and initializing them in the constructor

src/testing/time_refinement_study_advection.h
```diff
@@ -0,0 +1,47 @@
...
+    
+    /// Reinitialize parameters while refining the timestep. Necessary because all_parameters is constant.
+    Parameters::AllParameters reinit_params_and_refine_timestep(int refinement) const;
+
+
+    /// Parameters for the current refinement (i.e., duplicate of .prm file except for timestep size)
+    //Parameters::AllParameters params;
```
jbrillon, line 41: remove

tests/integration_tests_control_files/time_refinement_study/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+# =======================================
+# Time Study (Linear Advection Explicit RK)
+# =======================================
+# ----------------------------------------
+# - details: [if needed]
+# ----------------------------------------
+configure_file(time_study_advection_explicit.prm time_study_advection_explicit.prm COPYONLY)
+add_test(
+  NAME 1D_Time_Study_Advection_Explicit
```
jbrillon, line 11: consider making this all caps, seems to be our convention

src/physics/exact_solutions/exact_solution.cpp
```diff
@@ -0,0 +1,95 @@
...
+        : ExactSolutionFunction<dim,nstate,real>()
+{
+    t = 0;//time_compare;
+}
+
+template <int dim, int nstate, typename real>
+inline real ExactSolutionFunction_Zero<dim,nstate,real>
+::value(const dealii::Point<dim,real> &/*point*/, const unsigned int /*istate*/) const
+{
+    //NEED TO UPDATE THIS
```
dougshidong, line 21: Delete comment.

src/physics/exact_solutions/exact_solution.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+// ========================================================
+// ZERO -- Returns zero everywhere; used a placeholder when no exact solution is defined.
+// ========================================================
+template <int dim, int nstate, typename real>
+ExactSolutionFunction_Zero<dim,nstate,real>
+::ExactSolutionFunction_Zero()
+        : ExactSolutionFunction<dim,nstate,real>()
+{
+    t = 0;//time_compare;
```
dougshidong, line 14: Maybe use existing dealii::Function get_time, set_time, advance_time functions?

src/physics/exact_solutions/exact_solution.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    // Get the flow case type
+    const FlowCaseEnum flow_type = param->flow_solver_param.flow_case_type;
+    if (flow_type == FlowCaseEnum::advection_explicit_time_study) {
+        if constexpr (dim==1 && nstate==dim)  return std::make_shared<ExactSolutionFunction_1DSine<dim,nstate,real> > (time_compare);
+    } else {
+        // Select zero function if there is no exact solution defined
+        return std::make_shared<ExactSolutionFunction_Zero<dim,nstate,real>> ();
+    }
+    (void) time_compare; //to avoid
```
dougshidong, line 81: Delete unused line since it's being used now.

src/physics/exact_solutions/exact_solution.cpp
```diff
@@ -0,0 +1,95 @@
...
+::ExactSolutionFunction ()
+    : dealii::Function<dim,real>(nstate)//,0.0) // 0.0 denotes initial time (t=0)
+{
+    //time_compare +=0;
+}
+
+template <int dim, int nstate, typename real>
+std::shared_ptr<ExactSolutionFunction<dim, nstate, real>>
+ExactSolutionFactory<dim,nstate, real>::create_ExactSolutionFunction(
+    Parameters::AllParameters const *const param)
```
dougshidong, line 67: Don't pass AllParameters if FlowSolverParam is enough.

src/testing/flow_solver.cpp
```diff
@@ -323,6 +326,41 @@ void FlowSolver<dim,nstate>::output_restart_files(
...
+                 dealii::QGauss<dim>(poly_degree+10), //overintegrating by 10
+                 dealii::VectorTools::L2_norm);
+        double L2_error = dealii::VectorTools::compute_global_error(*dg->triangulation,
+                     difference_per_cell,
+                     dealii::VectorTools::L2_norm);
+        std::cout << "Computed error is " << L2_error << std::endl;
+        return L2_error;
+    }
+
+    //use deal.ii overintegration thing to calculate L2 error
```
dougshidong, line 357: thing?

src/physics/exact_solutions/exact_solution.cpp
jbrillon, line 21: clean up comments like this, can be confusing for reviewers

src/testing/flow_solver.cpp
```diff
@@ -372,7 +410,7 @@ int FlowSolver<dim,nstate>::run_test() const
         // Time advancement loop with on-the-fly post-processing
         //----------------------------------------------------
         pcout << "Advancing solution in time... " << std::endl;
-        while(ode_solver->current_time < final_time)
+        while((final_time - ode_solver->current_time) > (0.5 * constant_time_step)) //comparing to constant_time_step in case roundoff errors exceed machine zero
```
dougshidong, line 413: Instead of a while loop, would it make sense to somehow make sure that the last step lands exactly on the desired final time? Like if we want to go up to final_time = 1.0, but with constant time step of 0.3, we'd do
t0 = 0.0
t1 = t0 + 0.3 = 0.3
t2 = t1 + 0.3 = 0.6
t3 = t2 + 0.3 = 0.9
**t3 = t3 + 0.1 = 1.0**

Something like
```cpp
int n_steps = ((final_time - initial_time) / constant_step) // Rounds down to nearest integer
const time_remainder = final_time - n_steps * constant_time_step;
for (int i_step = 0; i_step < n_steps +1; ++i_step) 
{ 
    do work 
}
advance_solution_time(time_remainder);

```

@jbrillon

src/physics/exact_solutions/exact_solution.cpp
```diff
@@ -0,0 +1,95 @@
...
+ExactSolutionFunction_1DSine<dim,nstate,real>
+::ExactSolutionFunction_1DSine (double time_compare)
+        : ExactSolutionFunction<dim,nstate,real>()
+{
+    t = time_compare;
+}
+
+template <int dim, int nstate, typename real>
+inline real ExactSolutionFunction_1DSine<dim,nstate,real>
+::value(const dealii::Point<dim,real> &point, const unsigned int /*istate*/) const
```
jbrillon, line 40: I think `t` should be made an argument here rather than being fixed, and in your integration test you set `time_compare` and pass that there.

src/testing/flow_solver.cpp
dougshidong: Otherwise, we might end up comparing various solutions at different final times, which can easily mess up convergence orders.

src/parameters/parameters_flow_solver.cpp
```diff
@@ -94,6 +96,7 @@ void FlowSolverParam::parse_parameters(dealii::ParameterHandler &prm)
         else if (flow_case_type_string == "burgers_viscous_snapshot")  {flow_case_type = burgers_viscous_snapshot;}
         else if (flow_case_type_string == "burgers_rewienski_snapshot")  {flow_case_type = burgers_rewienski_snapshot;}
         else if (flow_case_type_string == "naca0012")  {flow_case_type = naca0012;}
+        else if (flow_case_type_string == "advection_explicit_time_study")  {flow_case_type = advection_explicit_time_study;}
```
jbrillon, line 99: Time study should not be a flow type, `advection` makes more sense, the purpose for `flow_case` is to only simulate the flow (integrate to the final time). The integration test `.cpp` file takes care of the `time_study` component, and the parameter file `.prm` and `ctest` name takes care of the `explicit` component

src/testing/flow_solver.cpp
dougshidong: OK, but the comment above still holds. What happens if constant_time_step is not an integer multiple of final time?

src/physics/initial_conditions/initial_condition.cpp
```diff
@@ -126,6 +126,30 @@ ::value(const dealii::Point<dim,real> &point, const unsigned int /*istate*/) con
...
+}
+
+template <int dim, int nstate, typename real>
+inline real InitialConditionFunction_1DSine<dim,nstate,real>
+::value(const dealii::Point<dim,real> &point, const unsigned int /*istate*/) const
+{
+    real value = 0;
+    real pi = dealii::numbers::PI;
+    if(point[0] >= 0.0 && point[0] <= 2.0){
+        value = sin(2*pi*point[0]/2.0);
```
jbrillon, line 147: if this is essentially the respective exact solution at `t=0`, consider calling it here and passing `t=0`

src/testing/flow_solver.cpp
jbrillon: For unsteady simulations that use a CFL number to compute the dt or `constant_time_step`, the final time is most likely not going to be exactly equal to the final time you see on the plots in papers

dougshidong: OK, but you would want to have the final time of the simulation to be the final time you describe in your paper... Need something such that if (current_time + time_step_evaluated_by_CFL_or_given_by_constant_time_step) is greater than (final_time), take a time_step = (final_time - current_time)

jbrillon: If you're using a CFL, I think its understood that when you write final time of the simulation, the actual final time could be off by a very small amount say 1e-4 or even smaller than that. Also, the current implementation ensures that we reach the _at least_ final time since it checks `current_time<final_time` then does `step_in_time` (which updates the current time) so the actual final time >= final time

dougshidong: Sure, so... you might be off up to O(dt)... Aka.. you're back to a Euler-explicit scheme accuracy...

src/testing/flow_solver_cases/periodic_1D_flow.cpp
cpethrick: I will move things into `periodic_cube_flow.cpp` (or renamed) after I've addressed other comments.

Mesh has been moved to `straight_periodic_cube.cpp`. I also added 2D square domain to make it complete.

src/testing/flow_solver_cases/periodic_1D_flow.cpp
cpethrick: Fixed in next commit.

src/testing/flow_solver.cpp
jbrillon: Okay sure, if I was comparing the solution at the final time then yes, but in that case, I would modify the CFL to be slightly different than the inputted one so that I exactly reach the inputted final time (done before advancing in time), I could introduce a function called `modify_time_step_to_exactly_reach_desired_final_time` and call that after doing `get_constant_time_step`; I can open a quick PR for this. What do you think?

src/testing/time_refinement_study_advection.cpp
cpethrick: That is a much cleaner solution. I was referencing `fd_state_sensitivity_wrt_parameter.cpp` when writing this section. Will fix in next commit.

src/mesh/grids/straight_periodic_cube.cpp
```diff
@@ -24,24 +24,37 @@ void straight_periodic_cube(std::shared_ptr<TriangulationType> &grid,
 
     // Definition for each type of grid
     std::string grid_type_string;
-    if(dim==3) {
+    const bool colorize = true;
+    dealii::GridGenerator::hyper_cube(*grid, domain_left, domain_right, colorize);
+    if (dim == 1){
```
jbrillon, line 29: use `if constexpr(dim == 1)` and similarly for the other ones since this is a template parameter

src/testing/time_refinement_study_advection.cpp
cpethrick: Forgot to remove that, it won't be used. Will fix in next commit.

src/testing/flow_solver_cases/periodic_1D_flow.cpp
jbrillon: Ok I'll resolve it once its all complete

src/testing/flow_solver_cases/periodic_1D_flow.cpp
jbrillon: Please only mark as fixed after the commit has been pushed

tests/integration_tests_control_files/time_refinement_study/CMakeLists.txt
cpethrick: Sure, I will change that and follow that convention in the future.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: Fixed comments in this file.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: Adjusted things so I don't need this line even if the zero function is selected.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: You're right, I'm only using parameters in the flow solver section. How do I pass only FlowSolverParam?

src/parameters/parameters_flow_solver.cpp
cpethrick: Thanks for the naming suggestion, I wasn't sure that this was the best choice! I would prefer to change the name to `advection_periodic` so that it isn't identical to the `pde_type`. I think this choice still reflects the contents of the `flow_case`, rather than the test that I'm running.

src/mesh/grids/straight_periodic_cube.cpp
cpethrick: Okay, fixing in the next commit.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: Yes, I will be more careful about that in the future.

src/testing/flow_solver.cpp
cpethrick: This was a note to  self - sorry for leaving it in. I'll be more careful about cleaning up comments in the future.

src/testing/flow_solver.cpp
cpethrick: I made this change because I'm using a `dt` which should evenly divide `final_time`. However, sometimes the computation would run until (`final_time + dt`) rather than stopping at `final_time`. The while loop entered because `current_time` was on the order of 1E-14 less than `final_time`. My fix compares to `dt` such that machine errors like this don't cause the flow solver to overshoot `final_time`. I don't have a particularly good reason for comparing to `0.5*dt` rather than any other constant of the same order.

I could revert this change and compare to a different `final_time_actual` (or something like that). However, I don't think this is the best solution going forward. Comparing at a different `final_time_actual` doesn't make a huge difference for advection. but for a more complicated PDE where the solution evolves between `final_time` and `final_time_actual`, I am not confident in the validity of comparing convergence at a different point in the solution. 

So, I'm not sure that my solution is the best one, but I think it's worth looking into a better way to handle this in the future.

src/testing/flow_solver_cases/periodic_1D_flow.cpp
jbrillon: Btw I think we should just leave it named as `PeriodicCubeFlow` since we're enforcing all domain sides to be of equal length

src/testing/flow_solver.cpp
cpethrick: I was debating whether this should be in `time_refinement_study` or `flow_solver`, and my logic in putting it here was that it would possibly useful for other tests, therefore putting it in `flow_solver` would make it easier to re-use.

If I were to move this to `time_refinement_study`, I would no longer need to store `exact_solution_function` in the flow solver. This might make it a little easier to ensure that consistent times are compared.

I would like to continue the discussion on this - not sure what is the better way to structure things to make `flow_solver` general but not overcomplicated!

src/physics/initial_conditions/initial_condition.cpp
cpethrick: Yes, it is. I think that falls under the discussion of should we unify exact solution and initial condition.

cpethrick:
The most recent commit fixes any minor comments. I've marked conversations that are quick fixes as resolved.

I think more discussion about the initial condition / exact solution / manufactured solution redundancy is needed. I'll summarize the discussion from the PR comments:

- Could make a unified "analytical function" to replace the three aforementioned analytical functions. 
- The `exact_solution` function as I've implemented it should take time as an argument rather than as a constant. This would be necessary (or at least useful) for comparing solutions that reach a different `final_time` than the target. 
- As suggested by Doug, I will look into dealii::Function get_time, set_time, advance_time to see if they make exact_solution better.

I suppose that the unified "analytical function" would not be in the scope of this PR. Therefore, should I remove exact_solution and group the solution into initial_condition, despite the naming? Or leave exact_solution as I've implemented it as temporary functionality?

src/testing/flow_solver.cpp
cpethrick: As a fix for this problem, my latest commit compares to 1E-13 rather than 0.5 dt to avoid taking an extra step. My test runs as before.

In the future, could look into making the tolerance scale with dt if this issue resurfaces, but this fix works with my test and should not impact the results of other tests.

src/physics/exact_solutions/exact_solution.cpp
dougshidong: The function signature would be `create_ExactSolutionFunction(const FlowSolverParam& flow_solver_parameters)`

And the calling function would call it like `create_ExactSolutionFunction(all_parameters->flow_solver_param)`

src/testing/flow_solver.cpp
dougshidong: Mmmmh, I still think that something like what I mentionned above is needed. I don't think you want the code to modify the input timestep or CFL. Since the code itself is a research code and we end up doing temporal studies, you'd want the specified timestep to be respected.
The solution I proposed with taking as many steps as you can until you need to reduce the last one is still my preferred solution.

This one:
t0 = 0.0
t1 = t0 + 0.3 = 0.3
t2 = t1 + 0.3 = 0.6
t3 = t2 + 0.3 = 0.9
**t4 = t3 + 0.1 = 1.0**

In effect, this would have a convergence of order O(0.3^p) even if the last step is 0.1, because most of the steps would be 0.3. If you decide to modify the stepping scheme, you would end up modifying it to 0.25, which wouldn't give the correct orders if you assumed it was halved.

dougshidong:
> The most recent commit fixes any minor comments. I've marked conversations that are quick fixes as resolved.
> 
> I think more discussion about the initial condition / exact solution / manufactured solution redundancy is needed. I'll summarize the discussion from the PR comments:
> 
> * Could make a unified "analytical function" to replace the three aforementioned analytical functions.
> * The `exact_solution` function as I've implemented it should take time as an argument rather than as a constant. This would be necessary (or at least useful) for comparing solutions that reach a different `final_time` than the target.
> * As suggested by Doug, I will look into dealii::Function get_time, set_time, advance_time to see if they make exact_solution better.
> 
> I suppose that the unified "analytical function" would not be in the scope of this PR. Therefore, should I remove exact_solution and group the solution into initial_condition, despite the naming? Or leave exact_solution as I've implemented it as temporary functionality?

I'm fine with not unifying the three aforementionned classes. It does become out of scope. Though, please do open an issue an link up that comment: https://github.com/dougshidong/PHiLiP/pull/138#issuecomment-1118967390

Pretty sure that using deal.II get_time, set_time, advance_time will solve that time input parameter stuff. You basically just have to call exact_solution.set_time(t) before calling exact_solution.value(t), and use const double t = get_time(), inside of value().

src/testing/time_refinement_study_advection.cpp
jbrillon: Ok sounds good, just mark comments as "fixed" once the commit has actually been pushed, helps make the review easier

src/testing/flow_solver_cases/periodic_1D_flow.cpp
jbrillon: In the future, let the reviewers mark comments as resolved, allows us to actually check the change and approve it or further comment on it

src/testing/flow_solver.cpp
jbrillon: Yeah I think it's better to move both `calculate_L2_error_at_final_time_wrt_function()` and `exact_solution_function` to `time_refinement_study`

src/testing/time_refinement_study_advection.cpp
```diff
@@ -0,0 +1,91 @@
...
+        }
+        L2_error_old = L2_error;
+    }
+
+    //Printing and writing convergence table
+    pcout << std::endl;
+    convergence_table.write_text(std::cout); //pcout gives an error. Shouldn't be an issue as this is 1D and doesn't use MPI
+
+    std::ofstream conv_tab_file;
+    const char fname[25] = "convergence_table_1D.txt";
```
jbrillon, line 79:
```suggestion
    const std::string fname = "temporal_convergence_table.txt";
```

jbrillon: want to keep things as general as possible

src/testing/time_refinement_study_advection.cpp
```diff
@@ -0,0 +1,91 @@
...
+                testfail = 1;
+                pcout << "Expected convergence order was not reached at refinement " << refinement <<std::endl;
+            }
+        }
+        L2_error_old = L2_error;
+    }
+
+    //Printing and writing convergence table
+    pcout << std::endl;
+    convergence_table.write_text(std::cout); //pcout gives an error. Shouldn't be an issue as this is 1D and doesn't use MPI
```
jbrillon, line 76: Take a look at `src/testing/flow_solvers_cases/periodic_cube_flow.cpp` function `compute_unsteady_data_and_write_to_table()` to see how it's being done using `pcout` and `mpi_rank`

src/testing/time_refinement_study_advection.cpp
cpethrick: Yes, it's general as long as there is an analytical solution available. Changing in the next commit to `TimeRefinementStudy`.

src/testing/flow_solver.h
```diff
@@ -26,7 +26,7 @@
 #include "ode_solver/explicit_ode_solver.h"
 #include "ode_solver/ode_solver_factory.h"
 #include "flow_solver_cases/periodic_cube_flow.h"
-#include "flow_solver_cases/periodic_1D_flow.h"
+//#include "flow_solver_cases/periodic_1D_flow.h"
```
jbrillon, line 29: delete

src/testing/flow_solver_cases/periodic_cube_flow.cpp
```diff
@@ -210,9 +210,41 @@ double PeriodicTurbulence<dim, nstate>::compute_kinetic_energy(DGBase<dim, doubl
...
+        const std::shared_ptr <dealii::TableHandler> /* unsteady_data_table */) const
+{
+    double dt = this->all_param.ode_solver_param.initial_time_step;
+    int output_solution_every_n_iterations = (int) (this->all_param.ode_solver_param.output_solution_every_dt_time_intervals/dt);
+ 
+    if ((current_iteration % output_solution_every_n_iterations) == 0){
+        this->pcout << "    Iter: " << current_iteration
+                    << "    Time: " << current_time
+                    << std::endl;
+    }
```
jbrillon, line 237: Instead of creating this class for the sole purpose of printing iteration and time, this should probably be done in the ODE solver `step_in_time` function. 

After line 55 here: https://github.com/dougshidong/PHiLiP/blob/1bed06d44bd9bcae1733c92efbca7b313262d81d/src/ode_solver/explicit_ode_solver.cpp#L55 

You could add:
`if ((this->ode_param.ode_output) == Parameters::OutputEnum::verbose &&
            (this->current_iteration%ode_param.print_iteration_modulo) == 0 ) {
           this->pcout << "    Iter: " << this->current_iteration
                                << "    Time: " << this->current_time
                                << std::endl;
}`

and use those parameters, or what ever makes most sense, in your `.prm` file.

src/mesh/grids/straight_periodic_cube.cpp
```diff
@@ -24,24 +24,37 @@ void straight_periodic_cube(std::shared_ptr<TriangulationType> &grid,
...
         std::vector<dealii::GridTools::PeriodicFacePair<typename dealii::Triangulation<dim>::cell_iterator> > matched_pairs;
         dealii::GridTools::collect_periodic_faces(*grid,0,1,0,matched_pairs);
         dealii::GridTools::collect_periodic_faces(*grid,2,3,1,matched_pairs);
         dealii::GridTools::collect_periodic_faces(*grid,4,5,2,matched_pairs);
         grid->add_periodicity(matched_pairs);
-        grid->refine_global(number_of_refinements);
     }
     else {
-    	std::cout << " ERROR: straight_periodic_cube() cannot be called for dim!=3 " << std::endl;
+    	std::cout << " ERROR: straight_periodic_cube() cannot be called for dim>3 " << std::endl;
```
jbrillon, line 49: Not sure this else statement is necessary anymore since `dim` will never be `>3`

src/testing/tests.cpp
```diff
@@ -169,6 +170,8 @@ ::select_test(const AllParam *const parameters_input,
         if constexpr (dim==3 && nstate==dim+2) return std::make_unique<TaylorGreenVortexEnergyCheck<dim,nstate>>(parameters_input,parameter_handler_input);
     } else if(test_type == Test_enum::taylor_green_vortex_restart_check) {
         if constexpr (dim==3 && nstate==dim+2) return std::make_unique<TaylorGreenVortexRestartCheck<dim,nstate>>(parameters_input,parameter_handler_input);
+    } else if(test_type == Test_enum::time_refinement_study_advection) {
```
jbrillon, line 173: rename to `time_refinement_study` to be consistent with the class name since the advection part is handled by the `.prm` file and denoted in the ctest name

src/testing/time_refinement_study_advection.cpp
```diff
@@ -0,0 +1,91 @@
+#include "time_refinement_study_advection.h"
```
jbrillon, line 1: rename .cpp/.h files to be consistent with the class name

tests/integration_tests_control_files/time_refinement_study/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+# =======================================
+# Time Study (Linear Advection Explicit RK)
+# =======================================
+# ----------------------------------------
+# - details: [if needed]
+# ----------------------------------------
+configure_file(time_study_advection_explicit.prm time_study_advection_explicit.prm COPYONLY)
+add_test(
+    NAME 1D_TIME_STUDY_ADVECTION_EXPLICIT
```
jbrillon, line 11: `1D_TEMPORAL_CONVERGENCE_STUDY_ADVECTION_EXPLICIT` instead?

src/parameters/parameters_flow_solver.cpp
```diff
@@ -94,6 +96,7 @@ void FlowSolverParam::parse_parameters(dealii::ParameterHandler &prm)
         else if (flow_case_type_string == "burgers_viscous_snapshot")  {flow_case_type = burgers_viscous_snapshot;}
         else if (flow_case_type_string == "burgers_rewienski_snapshot")  {flow_case_type = burgers_rewienski_snapshot;}
         else if (flow_case_type_string == "naca0012")  {flow_case_type = naca0012;}
+        else if (flow_case_type_string == "advection_periodic")  {flow_case_type = advection_periodic;}
```
jbrillon, line 99: add this here as well: 
https://github.com/dougshidong/PHiLiP/blob/1bed06d44bd9bcae1733c92efbca7b313262d81d/src/testing/flow_solver_cases/flow_solver_case_base.cpp#L37-L49

src/testing/flow_solver.cpp
cpethrick: @dougshidong I agree that your solution should preserve the order, because the remainder step is always smaller than the target dt, so error should still be dominated by (target_dt)^p. However, it's probably more sound for the time refinement studies to always use an evenly-divisible dt when doing temporal convergence studies. If the step size is calculated by a CFL number, then the final time could be adjusted at the first refinement (e.g. with your example, stop all computations at 0.9).

To address this, I will add an error check to the time refinement study that makes sure that `final_time` is divisible by `initial_time_step`. This will ensure that such an issue isn't encountered for a constant-time-step test like the one I have added.

Another complication to this discussion is the assumption that we know the step size a priori. I think that the solution I implemented a few commits ago will give me problems when I add relaxation RK timestepping. In that type of timestepping, the timestep taken at each step varies like dt_n = (1 + target_dt^(p-1) ) * target_dt. In that case, the 1E-13 tolerance is not guaranteed to capture the desired final time if it happens that the variations are on the lower side. @dougshidong's solution doesn't quite allow for this because it requires that we know `time_remainder` exactly. This isn't immediately necessary to be addressed, but an issue that I am anticipating for the future.

src/testing/flow_solver.cpp
cpethrick: After more thought, I agree. I've moved it in the most recent commit. Thanks for the suggestion!

src/testing/flow_solver.cpp
dougshidong: I think it's worth opening up an issue about it since we do care about OOA in a high order code like PHiLiP and it is fair that an unknowing student would assume that the final time specified actually is the final time reached.

src/testing/time_refinement_study_advection.cpp
cpethrick: Thanks for directing me to that similar function. The following fix behaves the same way (1D, no MPI):

`if (this->mpi_rank==0) convergence_table.write_text(std::cout); `

Is this what you meant? Passing `pcout` as the output stream gives an error:

`error: cannot convert ‘const dealii::ConditionalOStream’ to ‘std::ostream&’`

src/testing/flow_solver.h
cpethrick: Fixed in a recent commit

src/testing/flow_solver_cases/periodic_cube_flow.cpp
cpethrick: I could do that for flow_solver cases. However, that would make cases which use `advance_solution_time` print the current iteration twice due to [this](https://github.com/dougshidong/PHiLiP/blob/1bed06d44bd9bcae1733c92efbca7b313262d81d/src/ode_solver/ode_solver_base.cpp#L229) in `ode_solver_base.cpp`. In theory, I could add another statement to check whether it's a flow solver case, but I still feel like that is repeated functionality between `advance_solution_time` and 'step_in_time`. Also, I'll soon be adding a 1D periodic time refinement test which needs to track unsteady data. Therefore, while this class doesn't do much now, it is useful to keep here for future generality.

src/mesh/grids/straight_periodic_cube.cpp
cpethrick: Removing in next commit.

src/testing/tests.cpp
cpethrick: Renaming to `time_refinement_study` in next commit.

src/testing/time_refinement_study_advection.cpp
cpethrick: Renaming to `time_refinement_study` in next commit.

tests/integration_tests_control_files/time_refinement_study/CMakeLists.txt
cpethrick: Renaming to `1D_TIME_REFINEMENT_STUDY_ADVECTION_EXPLICIT` - I've been using the naming "time refinement" and it makes sense to be consistent. Will also change the .prm file name to match.

src/parameters/parameters_flow_solver.cpp
cpethrick: Adding in next commit.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: Thanks, I've changed to only pass FlowSolverParam.

src/testing/time_refinement_study_advection.h
cpethrick: Changed in most recent commit.

sivanadarajah:
> Wondered if there was a need for `physics/exact_solutions` as per previous discussion on Alex's PR about adding a time component to either manufactured solutions or initial conditions. Also, manufactured solutions and initial conditions are already very similar in that they basically just override the dealii::Function which already has a way to modify a time component
> 
> https://www.dealii.org/developer/doxygen/deal.II/classFunction.html https://www.dealii.org/developer/doxygen/deal.II/classFunctionTime.html#a3a583fd8f30db3549dbaa43b2592e2bc
> 
> At this point, I'm wondering if we even need to have different classes such as ManufacturedSolution, InitialConditions, and ExactSolution, and if we should instead just have a single SolutionFunctionFactory for all three (especially since some initial conditions are just the exact solutions at time t = 0).

I would

sivanadarajah:
> Wondered if there was a need for `physics/exact_solutions` as per previous discussion on Alex's PR about adding a time component to either manufactured solutions or initial conditions. Also, manufactured solutions and initial conditions are already very similar in that they basically just override the dealii::Function which already has a way to modify a time component
> 
> https://www.dealii.org/developer/doxygen/deal.II/classFunction.html https://www.dealii.org/developer/doxygen/deal.II/classFunctionTime.html#a3a583fd8f30db3549dbaa43b2592e2bc
> 
> At this point, I'm wondering if we even need to have different classes such as ManufacturedSolution, InitialConditions, and ExactSolution, and if we should instead just have a single SolutionFunctionFactory for all three (especially since some initial conditions are just the exact solutions at time t = 0).

You would still want to be able to have a different manufactured solutions then the prescribed initial condition. Wouldn't you loose that capability if you have a single SolutionFunctionFactory?

src/testing/flow_solver_cases/periodic_cube_flow.cpp
jbrillon: Good call, maybe just rename this class as `AdvectionPeriodic` then since its confusing having `PeriodicCubeFlow` and a `Periodic1DFlow` when cube includes them all

dougshidong:
> > Wondered if there was a need for `physics/exact_solutions` as per previous discussion on Alex's PR about adding a time component to either manufactured solutions or initial conditions. Also, manufactured solutions and initial conditions are already very similar in that they basically just override the dealii::Function which already has a way to modify a time component
> > https://www.dealii.org/developer/doxygen/deal.II/classFunction.html https://www.dealii.org/developer/doxygen/deal.II/classFunctionTime.html#a3a583fd8f30db3549dbaa43b2592e2bc
> > At this point, I'm wondering if we even need to have different classes such as ManufacturedSolution, InitialConditions, and ExactSolution, and if we should instead just have a single SolutionFunctionFactory for all three (especially since some initial conditions are just the exact solutions at time t = 0).
> 
> You would still want to be able to have a different manufactured solutions then the prescribed initial condition. Wouldn't you loose that capability if you have a single SolutionFunctionFactory?

Not really, that `SolutionFunctionFactory` would be responsible for returning a dealii::Function, which can then be used to either initialize our initial conditions or even to compare the final solution.

```cpp
class ExponentialSolutionFunction : public dealii::Function
{
    real value(dealii::Point p) const override {
        // implementation
    }
    dealii::Tensor gradient(dealii::Point p) const override {
        // implementation
   }
}
class SineSolutionFunction : public dealii::Function
{
    real value(dealii::Point p) const override {
        // implementation
    }
    dealii::Tensor gradient(dealii::Point p) const override {
        // implementation
   }
}
class SolutionFunctionFactory
{
    std::shared_ptr<dealii::Function> createSolutionFunction(const SolutionFunctionType function_type)
    {
        switch(function_type) {
            case(SolutionFunctionType::ExponentialSolution) return std::make_shared<ExponentialSolutionFunction>(); break;
            case(SolutionFunctionType::SineSolution) return std::make_shared<SineSolutionFunction>(); break;
            // etc.
        }
    }
}
```
Then in the code, the resulting dealii::Function can be used in various ways:
```cpp
void test()
{
    // some initialization
    // WAY #1
    std::shared_ptr<dealii::Function> initial_conditions_function = createSolutionFunction(function_type)
    VectorTools::project ( mapping, dof, constraints, quadrature, 
                                       *initial_conditions_function, // used here
                                       dg->solution);
    // WAY #2
     std::shared_ptr<dealii::Function> manufactured_solution_function = createSolutionFunction(other_function_type);
    // Somehow use manufactured_solution_function to define source terms within Physics/DG

    // Some other code to run the solver

    // WAY #3
    std::shared_ptr<dealii::Function> exact_solution = createSolutionFunction(another_function_type_again);
    // Code to compare final dg->solution to exact_solution
    
```

Being able to update the time parameter is already a feature of the dealii::Function. Right now, our ManufacturedSolution, InitialConditions, and ExactSolution, are a layer of abstraction that don't really add to the code in terms of functionality since the implementations are in another layer of derived classes.

We basically have:
............................/---- ManufacturedSolution --- Bunch of different ManufacturedSolutions
dealii::Function --|---- InitialConditions           --- Bunch of different InitialConditions
............................ \ ---- ExactSolution              --- Bunch of different ExactSolution

But the ManufacturedSolution, InitialConditions, ExactSolution, have the exact same code.

src/testing/flow_solver.cpp
```diff
@@ -6,6 +6,8 @@
 #include <stdlib.h>
 #include <vector>
 #include <sstream>
+#include <deal.II/numerics/vector_tools.h>
```
jbrillon, line 9: no longer needed

src/testing/time_refinement_study.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    double final_time = this->all_parameters->flow_solver_param.final_time;
+    double initial_time_step = this->all_parameters->ode_solver_param.initial_time_step;
+    int n_steps = round(final_time/initial_time_step);
+    if (n_steps * initial_time_step != final_time){
+        pcout << "Error: final_time is not evenly divisible by initial_time_step!" << std::endl
+              << "Remainder is " << fmod(final_time, initial_time_step)
+              << ". Modify parameters to run this test." << std::endl;
+        std::abort();
+    }
```
jbrillon, line 69: given this constraint, is the following change `while((ode_solver->current_time) < (final_time - 1E-13))` in `flow_solver.cpp` still necessary?

src/testing/time_refinement_study.cpp
```diff
@@ -0,0 +1,130 @@
...
+    double order_tolerance = 0.1;
+
+    dealii::ConvergenceTable convergence_table;
+    double L2_error_old = 0;
+    double L2_error_conv_rate=0;
+
+    for (int refinement = 0; refinement < n_time_calculations; ++refinement){
+        
+        pcout << "\n\n---------------------------------------------\n Refinement number " << refinement <<
+            " of " << n_time_calculations - 1 << std::endl << "---------------------------------------------" << std::endl;
```
jbrillon, line 82: clean this up into separate `pcout` statements

src/testing/time_refinement_study_advection.h
jbrillon: I still see these in the `time_refinement_study.h` file

src/testing/flow_solver.cpp
jbrillon: Ok, how about just outputting a warning if `mod(final_time,constant_time_step)!=0` ?

src/mesh/grids/straight_periodic_cube.cpp
```diff
@@ -24,24 +24,33 @@ void straight_periodic_cube(std::shared_ptr<TriangulationType> &grid,
...
-    }
-    else {
-    	std::cout << " ERROR: straight_periodic_cube() cannot be called for dim!=3 " << std::endl;
-    	std::abort();
     }
+    grid->refine_global(number_of_refinements);
 }
 
+template void straight_periodic_cube<1, dealii::Triangulation<1>> (std::shared_ptr<dealii::Triangulation<1>> &grid, const double domain_left, const double domain_right, const int number_of_cells_per_direction);
+template void straight_periodic_cube<2, dealii::parallel::distributed::Triangulation<2>> (std::shared_ptr<dealii::parallel::distributed::Triangulation<2>> &grid, const double domain_left, const double domain_right, const int number_of_cells_per_direction);
```
jbrillon, line 52: can group the two below lines with `#if PHILIP_DIM!=1` and passing `PHILIP_DIM` instead of `2` and `3`

src/testing/flow_solver_cases/periodic_cube_flow.cpp
jbrillon: And further move the `AdvectionPeriodic` class into to a separate file `flow_solver_cases/advection_periodic.cpp/.h`

jbrillon:
> 

Just to add when you say they have the "exact same code", the only difference is that the `ExactSolutionFunction` and `InitialConditionFunction` classes only require `value` to be defined by their derived classes, whereas `ManufacturedSolution` requires `value`, `gradient`, and `hessian` to be defined by the derived classes; so really we'd only be unifying the `value` member function aspect of the classes.

Unifying them would be good, at least unifying `InitialConditionFunction` and `ExactSolutionFunction` would be a good start

src/testing/time_refinement_study_advection.h
cpethrick: I saved that file incorrectly. Actually fixed now!

src/testing/time_refinement_study.cpp
cpethrick: Yes, because the actual final time that is reached is something like (final_time - 1E-14), so without that fix, it will calculate an additional step to (final_time - 1E-14 + dt). I'm assuming that this is due to floating point representation/machine error/etc.

src/testing/time_refinement_study.cpp
cpethrick: Okay, will modify.

src/testing/time_refinement_study_advection.h
jbrillon: They're still being assigned in the header file

src/testing/time_refinement_study.cpp
```diff
@@ -0,0 +1,130 @@
...
+        double L2_error = calculate_L2_error_at_final_time_wrt_function(flow_solver->dg, params);
+        pcout << "Computed error is " << L2_error << std::endl;
+
+        const double dt =  params.ode_solver_param.initial_time_step;
+        convergence_table.add_value("refinement", refinement);
+        convergence_table.add_value("dt", dt );
+        convergence_table.set_precision("dt", 3);
+        convergence_table.set_scientific("dt", true);
+        convergence_table.add_value("L2_error",L2_error);
+        convergence_table.set_precision("L2_error", 4);
```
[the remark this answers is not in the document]
cpethrick: Yes, good point. I'll change to all 16 digits for L2 error.

src/mesh/grids/straight_periodic_cube.cpp
cpethrick: Sure, that's a bit nicer. Changing in next commit.

src/testing/time_refinement_study_advection.h
cpethrick: Yes, the fix will be in my next commit.

src/testing/time_refinement_study_advection.cpp
cpethrick: I will add @dougshidong's fix in my next commit.

src/testing/flow_solver_cases/periodic_cube_flow.cpp
cpethrick: I am renaming to `Periodic1DUnsteady`, as this same flow solver setup will be used with Burgers' for a future time refinement study. I am also moving into separate files, `periodic_1D_unsteady.<cpp/h>`.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: I have changed `create_ExactSolutionFunction()` to take a time as input, which I think is sufficient for this PR's needs. I think those functions are probably useful to include in the future when working with the `dealii::Function` sections, though.

cpethrick:
My recent commits address all comments made on this PR, as well as merging in the most recent master. I will open an issue regarding the unification of the various `dealii::Function` codes and include the highlights of the discussion here.

dougshidong:
> My recent commits address all comments made on this PR, as well as merging in the most recent master. I will open an issue regarding the unification of the various `dealii::Function` codes and include the highlights of the discussion here.

Thanks. Will resolve and approve when the issue is opened.

dougshidong (approved):

src/physics/initial_conditions/initial_condition.h
```diff
@@ -151,6 +151,24 @@ class InitialConditionFunction_BurgersViscous: public InitialConditionFunction<d
...
+/// Initial Condition Function: 1D Sine Function; used for temporal convergence
+template <int dim, int nstate, typename real>
+class InitialConditionFunction_1DSine
+        : public InitialConditionFunction<dim,nstate,real>
+{
+protected:
+    using dealii::Function<dim,real>::value; ///< dealii::Function we are templating on
+
+public:
+    /// Constructor for InitialConditionFunction_BurgersRewienski
```
jbrillon, line 163: update comment for doxygen

src/mesh/grids/straight_periodic_cube.cpp
```diff
@@ -24,24 +24,36 @@ void straight_periodic_cube(std::shared_ptr<TriangulationType> &grid,
...
-    else {
-    	std::cout << " ERROR: straight_periodic_cube() cannot be called for dim!=3 " << std::endl;
-    	std::abort();
     }
+    grid->refine_global(number_of_refinements);
 }
 
-template void straight_periodic_cube<3, dealii::parallel::distributed::Triangulation<3>> (std::shared_ptr<dealii::parallel::distributed::Triangulation<3>> &grid, const double domain_left, const double domain_right, const int number_of_cells_per_direction);
+#if PHILIP_DIM==1
+    template void straight_periodic_cube<1, dealii::Triangulation<1>> (std::shared_ptr<dealii::Triangulation<1>> &grid, const double domain_left, const double domain_right, const int number_of_cells_per_direction);
```
jbrillon, line 52:
```suggestion
    template void straight_periodic_cube<PHILIP_DIM, dealii::Triangulation<PHILIP_DIM>> (std::shared_ptr<dealii::Triangulation<PHILIP_DIM>> &grid, const double domain_left, const double domain_right, const int number_of_cells_per_direction);
```

src/parameters/parameters_flow_solver.cpp
```diff
@@ -94,6 +96,17 @@ void FlowSolverParam::declare_parameters(dealii::ParameterHandler &prm)
...
+        prm.enter_subsection("time_refinement_study");
+        {
+            prm.declare_entry("number_of_times_to_solve", "4",
+                              dealii::Patterns::Integer(1, dealii::Patterns::Integer::max_int_value),
+                              "Number of times to run the flow solver during a time refinement study.");
+            prm.declare_entry("refinement_ratio", "0.5",
+                              dealii::Patterns::Double(0, 1.0),
+                              "Ratio between a timestep size and the next in a time refinement study, 0<r<1.");
+        }
+        prm.leave_subsection();
```
jbrillon, line 109: We'd probably want to move this into a file called `parameters_time_refinement_study.cpp` like we have for grid refinement and manufactured convergence studies since they are used by the test and not the flow solver, you can open an issue for it and address it later if you want -- we only want flow related parameters here

src/testing/flow_solver_cases/periodic_1D_unsteady.h
```diff
@@ -0,0 +1,50 @@
...
+#include <deal.II/distributed/tria.h>
+
+namespace PHiLiP {
+namespace Tests {
+
+#if PHILIP_DIM==1
+    using Triangulation = dealii::Triangulation<PHILIP_DIM>;
+#else
+    using Triangulation = dealii::parallel::distributed::Triangulation<PHILIP_DIM>;
+#endif
```
jbrillon, line 20: No need for this since the grid is handled by your parent class `PeriodicCubeFlow`

src/physics/exact_solutions/exact_solution.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+// ========================================================
+// ZERO -- Returns zero everywhere; used a placeholder when no exact solution is defined.
+// ========================================================
+template <int dim, int nstate, typename real>
+ExactSolutionFunction_Zero<dim,nstate,real>
+::ExactSolutionFunction_Zero(double time_compare)
+        : ExactSolutionFunction<dim,nstate,real>()
+{
+    t = time_compare;
```
donovan97, line 14: Can place in the member initializer list, as `t(time_compare)`

src/physics/exact_solutions/exact_solution.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+//=========================================================
+// FLOW SOLVER -- Exact Solution Base Class + Factory
+//=========================================================
+template <int dim, int nstate, typename real>
+ExactSolutionFunction<dim,nstate,real>
+::ExactSolutionFunction ()
+    : dealii::Function<dim,real>(nstate)
+{
+    //do nothing
```
donovan97, line 59: What is the purpose of the base class if it does nothing? If it is just to define a `ExactSolutionFunction` type, you might want to use an abstract class.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: Good point, that is a nicer way to initialize. I will change that in the next commit.

src/physics/exact_solutions/exact_solution.cpp
cpethrick: I used the same structure as `initial_condition.cpp`, which also uses an [empty base class](https://github.com/dougshidong/PHiLiP/blob/b13f5ff03c9ee050d21d1716516908e0b6513072/src/physics/initial_conditions/initial_condition.cpp#L164). I would rather leave this topic to be addressed with Issue #143.

src/parameters/parameters_flow_solver.cpp
cpethrick: I agree that that's a better way to organize things. I'll open an issue.

src/physics/exact_solutions/exact_solution.cpp
dougshidong: Yup, that's what I had previously noticed, it is basically a middle interface to dealii::Function that adds no functionality, just like InitialConditions. Issue #143 will remedy that.

src/physics/initial_conditions/initial_condition.h
cpethrick: Changed, as well as in `exact_solution.h`.

src/physics/exact_solutions/exact_solution.cpp
donovan97: Sounds good.

jbrillon (approved):

dougshidong:
Looks like everything got resolved or an issue has been opened. @jbrillon and @donovan97 , please resolve the comments and merge it in.

dougshidong:
Thanks again @cpethrick ! Bigger review than I expected, but it happens.

PR #139, opened 2022-05-06
Restructuring PeriodicCubeFlow

jbrillon:
PR to restructure `PeriodicCubeFlow` by deriving `PeriodicTurbulence` from this class; done in support of PR #138.

### DIFF
--- a/src/testing/flow_solver_cases/flow_solver_case_base.cpp
+++ b/src/testing/flow_solver_cases/flow_solver_case_base.cpp
@@ -14,7 +14,7 @@ FlowSolverCaseBase<dim, nstate>::FlowSolverCaseBase(const PHiLiP::Parameters::Al
 
 
 template <int dim, int nstate>
-void FlowSolverCaseBase<dim,nstate>::display_flow_solver_setup(std::shared_ptr<InitialConditionFunction<dim,nstate,double>> /*initial_condition*/) const
+void FlowSolverCaseBase<dim,nstate>::display_flow_solver_setup(std::shared_ptr<InitialConditionFunction<dim,nstate,double>> initial_condition) const
 {
     using PDE_enum = Parameters::AllParameters::PartialDifferentialEquation;
     const PDE_enum pde_type = all_param.pde_type;
@@ -26,6 +26,14 @@ void FlowSolverCaseBase<dim,nstate>::display_flow_solver_setup(std::shared_ptr<I
     pcout << "- PDE Type: " << pde_string << std::endl;
     pcout << "- Polynomial degree: " << this->all_param.grid_refinement_study_param.poly_degree << std::endl;
     pcout << "- Final time: " << this->all_param.flow_solver_param.final_time << std::endl;
+
+    this->display_additional_flow_case_specific_parameters(initial_condition);
+}
+
+template <int dim, int nstate>
+void FlowSolverCaseBase<dim,nstate>::display_additional_flow_case_specific_parameters(std::shared_ptr<InitialConditionFunction<dim,nstate,double>> /*initial_condition*/) const
+{
+    // Do nothing
 }
 
 template <int dim, int nstate>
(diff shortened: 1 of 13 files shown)


dougshidong:
Not sure I fully understand the goal of this commit and how it supports #138, but I see nothing "wrong" with this PR.

dougshidong (approved):

jbrillon:
@dougshidong https://github.com/dougshidong/PHiLiP/pull/138#discussion_r866233739

cpethrick (commented):
Thanks for restructuring this! I think deriving from PeriodicCubeFlow is a nice way to organize the flow solver cases, especially as we increase the number of cases. I left one comment on some functions that could be moved to PeriodicTurbulence.

dougshidong:
Aaah I see. It was some copy pasted code. Yes this PR makes sense to me now.

src/testing/flow_solver_cases/periodic_cube_flow.h
```diff
@@ -75,6 +57,36 @@ class PeriodicCubeFlow : public FlowSolverCaseBase<dim,nstate>
     double integrate_over_domain(DGBase<dim, double> &dg,const IntegratedQuantitiesEnum integrated_quantity) const;
```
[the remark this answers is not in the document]
jbrillon: fixed in most recent commit

src/testing/flow_solver_cases/flow_solver_case_base.cpp
```diff
@@ -26,6 +26,14 @@ void FlowSolverCaseBase<dim,nstate>::display_flow_solver_setup(std::shared_ptr<I
     pcout << "- PDE Type: " << pde_string << std::endl;
     pcout << "- Polynomial degree: " << this->all_param.grid_refinement_study_param.poly_degree << std::endl;
     pcout << "- Final time: " << this->all_param.flow_solver_param.final_time << std::endl;
+
+    this->display_additional_flow_case_specific_parameters(initial_condition);
```
donovan97, line 30: Why is `this` keyword being used in the base class, and why is it calling a function that does nothing?

src/testing/flow_solver_cases/flow_solver_case_base.cpp
```diff
@@ -26,6 +26,14 @@ void FlowSolverCaseBase<dim,nstate>::display_flow_solver_setup(std::shared_ptr<I
...
+}
+
+template <int dim, int nstate>
+void FlowSolverCaseBase<dim,nstate>::display_additional_flow_case_specific_parameters(std::shared_ptr<InitialConditionFunction<dim,nstate,double>> /*initial_condition*/) const
```
donovan97, line 34: Is this function really necessary? Why not just override display_flow_solver_setup and just display whatever you want for each case?

jbrillon: to avoid repeating code for things that should be displayed by default

src/testing/flow_solver_cases/flow_solver_case_base.cpp
jbrillon: Personal preference, I like it since using `this` explicitly marks the object members

jbrillon: For `display_additional_flow_case_specific_parameters`, see comment https://github.com/dougshidong/PHiLiP/pull/139#discussion_r867372889

src/testing/flow_solver_cases/flow_solver_case_base.cpp
donovan97: In that case, display_additional_flow_case_specific_parameters should probably be pure virtual and implemented in every class derived from flow_solver_case_base. `display_flow_solver_setup` should probably not be virtual if it is not meant to be overridden. Also, if you could add each pde_type to `display_flow_solver_setup`, that would probably be good since that will likely make a small bug when someone tries to add a new flow solver case with a new pde type.

jbrillon: Addressed in most recent commit

donovan97 (approved):

PR #140, opened 2022-05-08
fix compiler complaints

cyanmcg:
fix some errors from the compiler


cyanmcg:
It compiled fine on my system Fedora 35 with OpenMPI 4.1.3

Issue #120, opened 2022-04-05
FlowSolver standalone class


jbrillon:
Move `flow_solver.cpp` into its own directory. See discussion in PR #119 `src/testing/flow_solver.cpp`.

dougshidong:
https://github.com/dougshidong/PHiLiP/pull/119#discussion_r842306331

PR #147, opened 2022-06-08
FlowSolver standalone class

jbrillon:
Pull request to close issue #120. Introducing the `run_type` parameter for running either: `integration_test` or `flow_simulation`.


src/flow_solver/flow_solver.cpp
```diff
@@ -475,17 +480,47 @@ ::create_FlowSolver(const Parameters::AllParameters *const parameters_input,
     return nullptr;
 }
 
+template<int dim, int nstate>
+std::unique_ptr< FlowSolverBase > FlowSolverFactory<dim,nstate>
```
donovan97, line 484: FlowSolverFactory could be moved to a new file to match other factories in the code like for ODE and Physics. The FlowSolver class is getting big, so maybe its time they each get their own file.

donovan97 (commented):
Great PR, thanks for doing this. It's something I've been looking forward to for a while.

src/flow_solver/flow_solver.h
```diff
@@ -3,16 +3,15 @@
...
 #include "parameters/all_parameters.h"
 
 // for generate_grid
+#include <deal.II/grid/tria.h>
 #include <deal.II/distributed/shared_tria.h>
 #include <deal.II/distributed/tria.h>
 
-// for the grid:
-#include "grid_refinement_study.h"
+// #include "grid_refinement_study.h" // remove?
```
cpethrick, line 14: Commented line has not been removed. There's also a similar comment on line 36 of this file.

cpethrick (commented):
Great PR! This will make it much more convenient to use the flow_solver structure. The code looks good - I identified a commented line that should be removed but otherwise I didn't identify any issues.

One point of discussion: could we move the tests which use flow_solver into a subdirectory? E.g., `testing/flow_solver_tests`.  This would make it easier to distinguish the tests that use `flow_solver` from those that don't. What do you think?

src/flow_solver/flow_solver.h
jbrillon: fixed in most recent commit

jbrillon:
> 
Ideally, we would want all the older tests to take advantage of the flow_solver for taking care of getting the flow solution, greatly reducing code repetition, so that would mean they all fall under the `src/testing/` directory. I think its fine to leave it as is since we know that this is an existing issue we want to address in the future. Or we could introduce two subdirectories for reorganizing the (1) `src/testing/new_tests/`, and (2) `src/testing/old_tests/`; where we'd eventually want to move the old tests into the new tests once they've been changed to use flow solver. @dougshidong @donovan97 what do you think?

dougshidong:
Nice clean up.

The `testing` file and folder restructure is a known improvement to be done. Feel free to open an issue about it @cpethrick .

The current will go well with the issue I just opened for a future clean up using ctest labels:
https://github.com/dougshidong/PHiLiP/issues/148

One thing to keep in mind is that a flow simulation can also be an integration test. We need to be careful about pulling out tests out of the integration test suite even if it uses flow solver.

@cpethrick since you seem to have reviewed the code as well, please feel free to approve and merge it if you're satisfied with his answer. I'm trying to stop approving PR as much as possible and leaving it up to you all to approve each others PR.

cpethrick (approved):
I will open an issue regarding the organization of tests and migrating tests to use the flow solver. 

Other than that, I see no issues, so I will approve this PR and merge it in.

Issue #144, opened 2022-05-11
Move test-specific parameters out of parameters_flow_solver


cpethrick:
As per a [comment](https://github.com/dougshidong/PHiLiP/pull/138#discussion_r870406297) on PR #138,  move any integration test-specific parameters (i.e., those for the time refinement study or for the TGV integration tests) into individual files.

cpethrick:
Closing this -- time refinement params have been moved out and it is fine to leave the TGV params as-is because they are used to initialize the flow field.

PR #111, opened 2022-02-25
ESFR Split Formulation

AlexanderCicchino:
This pull request contains the ESFR split form. The changes in this pull request are as follows:
1) The DG Strong form explicit introduced: constructing metric terms on the fly along with the assembly of metric dependent operators on the fly. It enables uncollocated split-forms with respect to the curvilinear gradient/divergence operator assembled based on whether the gradient, divergence or hybrid of them both are to be used. Additionally, the operators class was used throughout to clearly distinguish between what is truly being interpolated on what, and all the necessary projections intertwined. A clear example of this would be on the surface terms with interpolating volume fluxes from volume cubature nodes to the facet. 
2) Consistently clear explicit references to what is physical and what is reference is performed throughout DG Strong. This is critical for provably stable curvilinear tests. An example being the use of a physical normal for the surface numerical flux, but a reference normal for the interpolation of the reference volume flux interpolated to the facet. 
3) In DG, the mass matrix now has ESFR capabilities through the operators class. Additionally, for curvilinear elements, the mass matrix has the option to use a weight-adjusted low-storage implementation for both DG and ESFR cases. All ESFR mass matrices are verified through the operators volume operators test.
4) The metric terms assembled via the operators class are provably free-stream preserving via GCL tests. Default is the conservative curl for, it also can use the invariant curl form by passing the paramter "use_invariant_curl_form." Both forms have unit tests for GCL in both the volume and surface, along with ensuring the surface metrics are conforming, and surface parametric GCL tests.
5) ODE explicit solver has Runge-Kutta 4 added to it. Also, ODE solver base has modifications for both orders of accuracy versus energy tests. The modification specifically is an if statement checking whether an energy test is used. This will be an issue to raise after the pull request to add the energy and conservation check within the ode class. 
6) Unsteady source terms were added, with DG tracking the current time. For a future issue, the unsteady source term should be introduced in the manufactured solution class with time dependence. In manufactured solution, there is an "Alex function": this was used from the lab group meeting where I showed how to use the manufactured solution. I will use this as a base reference for the unsteady time dependent source in the future issue.
7) Surface splitting functions were introduced in the physics classes.
8) Central numerical flux was added for stability tests for linear advection on curvilinear grids. Entropy conserving numerical flux added currently only for Burgers' equation. (Both these fluxes conserve entropy for their respective test cases)
9) Conservation and energy tests implemented for general FR problems in the unsteady tests. Example: src/testing/burgers_stability.cpp 
[...]

Issue #86, opened 2021-10-07
Test 1D_burgers_stability_LONG fails


donovan97:
The test used to pass when pseudotime was set to true in ODESolverBase::advance_solution_time, however this was found to be a bug and pseudotime should be false here. Fixing this bug causes the test to fail.

jbrillon:
Addressed by PR #111

PR #152, opened 2022-06-16
 Burgers' time refinement study and first implementation of relaxation Runge-Kutta

cpethrick:
This PR has two objectives:

1. Add a new flow solver test, `time_refinement_study_reference.<cpp/h>`. As indicated by the naming, this time refinement study is intended for tests where no analytical solution exists, therefore must be compared to a reference solution. The associated test, `1D_TIME_REFINEMENT_STUDY_BURGERS_EXPLICIT`, tests the convergence order of the inviscid Burgers' equation with a sinusoidal initial condition.

- This test is configured such that it can also use RRK by modifying the `.prm` file. In cases using RRK, the final time reached is generally not the same as the target final time. Therefore, a new reference solution is calculated if the target final time is not reached.
- Additional data is printed in a convergence table according to the test setup: energy if collocated nodes are used and relaxation factor gamma if RRK is used.
- `periodic_1D_unsteady.cpp` is modified to print energy in an unsteady data file. Once again, this is only done for collocated nodes according to the calculation method.
- Move parameters associated with the two time refinement studies from `parameters_flow_solver.<pp/h>` into a new file, `parameters_time_refinement_study.<cpp/h>`, according to Issue #144.

2. Add a first implementation of the [relaxation Runge-Kutta method](https://doi.org/10.1137/19M1263662). This implementation is only valid for cases where numerical entropy is energy (i.e., the inviscid Burgers' equation), and collocated nodes are used.

- The implementation is in a new ODE solver class in `rrk_explicit_ode_solver.<cpp/h>` which is derived from `ExplicitODESolver`
- A test is added to verify the implementation, `1D_BURGERS_ENERGY_CONSERVATION_RRK`. This test (based on `flow_solver`) checks energy conservation for four cases, (no RRK, RRK)x(large timestep, small timestep). Energy is expected to be conserved for all cases except (no RRK, large timestep) due to the split form and RRK.

Minor changes have also been made to keep the time refinement tests up-to-date with the changes to `flow_solver`.

**Note: Burgers' tests are currently failing.** This is due to the bug identified by issue #86, where a bug in deal.ii's 1D periodic boundaries cause energy to increase slightly. During development, I used a hard-coded fix by Alex (see changes in commit history [here](https://github.com/dougshidong/PHiLiP/pull/152/commits/0f6706cf94f928443495cc8c743d35bafec623e2), reverted to master [here](https://github.com/dougshidong/PHiLiP/pull/152/commits/a1b64fdbe1f05f0e729f8207d83029289c8256e8)). These tests should pass once PR #111 goes through. Without the changes in PR #111, these two tests accumulate a small amount of energy. This is enough to make `1D_BURGERS_ENERGY_CONSERVATION_RRK` fail, but not `1D_TIME_REFINEMENT_STUDY_BURGERS_EXPLICIT`.


cpethrick:
I am marking this PR as ready for review so that I can begin work off of @AlexanderCicchino's fork rather than waiting for PR #111 to be merged. Merging before ESFR is added means that one test I've introduced is currently failing because of the periodic boundary bug that is resolved by PR #111. 

As well, the implementation for RRK is currently only valid for collocated nodes due to the energy calculation method. I was hoping to generalize this after ESFR is added to PHiLiP. As it's difficult to do any validation for energy conservation on uncollocated nodes for now, I am leaving the implementation as-is, with the intention to move energy calculation into `physics/burgers.cpp`, in the same way that kinetic energy is calculated in `euler.cpp`.

src/testing/burgers_energy_conservation_rrk.cpp
```diff
@@ -0,0 +1,161 @@
...
+        return 1; //fail test
+    }else{
+        pcout << "Energy was conserved, but was expected NOT to be conserved." << std::endl;
+        pcout << "    Unexpected result! Test failing." << std::endl;
+        return 1; //fail test (included for completeness, but not expected to be used)
+    }
+}
+
+template <int dim, int nstate>
+double BurgersEnergyConservationRRK<dim, nstate>::compute_energy_collocated(
```
donovan97, line 60: Is there a reason why the compute_energy_collocated function is repeated in the tests even though the same function seems to exist in the Periodic1DUnsteady flow solver case?

donovan97 (commented):
Nice PR, great work!

dougshidong:
Please add the failing test to the issue list with the `testfail` label

https://github.com/dougshidong/PHiLiP/issues?q=is%3Aissue+is%3Aopen+label%3Atestfail

src/flow_solver/flow_solver_factory.cpp
```diff
@@ -46,7 +46,7 @@ ::select_flow_case(const Parameters::AllParameters *const parameters_input,
             std::shared_ptr<FlowSolverCaseBase<dim, nstate>> flow_solver_case = std::make_shared<NACA0012<dim,nstate>>(parameters_input);
             return std::make_unique<FlowSolver<dim,nstate>>(parameters_input, flow_solver_case, parameter_handler_input);
         }
-    } else if (flow_type == FlowCaseEnum::advection_periodic){
+    } else if ((flow_type == FlowCaseEnum::advection_periodic)||(flow_type == FlowCaseEnum::burgers_periodic)){
```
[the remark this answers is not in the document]
cpethrick: Good point. I think that the two cases are currently similar enough that it's okay to re-use Periodic1DUnsteady, and I don't intend on making too many more changes. But I'll keep an eye on the overlap and separate them if I need a lot more if statements.

tests/integration_tests_control_files/burgers_split_stability/CMakeLists.txt
```diff
@@ -5,3 +5,12 @@ add_test(
   COMMAND mpirun -n 1 ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_1D -i ${CMAKE_CURRENT_BINARY_DIR}/1D_burgers_stability.prm
   WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
 )
+
+
```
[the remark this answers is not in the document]
cpethrick: Good idea, I have cleaned up `CMakeLists.txt` in both folders I've put tests into.

src/testing/burgers_energy_conservation_rrk.cpp
cpethrick: No reason, I just didn't think to calculate energy through the flow solver case. I agree that calculating in Periodic1DUnsteady is better to reduce repetition and have changed in the most recent commit. Thanks for the suggestion!

cpethrick:
> Please add the failing test to the issue list with the `testfail` label
> 
> https://github.com/dougshidong/PHiLiP/issues?q=is%3Aissue+is%3Aopen+label%3Atestfail

I will do that right away and mention this PR. Thanks!

donovan97 (approved):

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
```diff
@@ -21,24 +11,74 @@ namespace FlowSolver {
...
+
+}
+
+
+template <int dim, int nstate>
+double Periodic1DUnsteady<dim, nstate>::compute_energy_collocated(
+        const std::shared_ptr <DGBase<dim, double>> dg
+        ) const
+{
+    // Intention is to eventually calculate energy in physics and generalize to arbitrary nodes
```
jbrillon, line 25: Fine for now, you could take a look at `periodic_turbulence.cpp` to see how its done

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -0,0 +1,109 @@
...
+    //assemble solution from stages
+    for (int i = 0; i < this->rk_order; ++i){
+       this->solution_update.add(
+               dt * relaxation_parameter * this->butcher_tableau_b[i],
+               this->rk_stage[i]); 
+    }
+    this->dg->solution = this->solution_update; // u_np1 = u_n +gamma*dt* sum(k_i * b_i)
+
+    ++(this->current_iteration);
+    this->current_time += relaxation_parameter * dt;
```
jbrillon, line 52: this seems to be the only difference between `ExplicitODESolver`'s `step_in_time()` function, consider defining a virtual function in `ExplicitODESolver` called `update_current_time` and by default would do `this->current_time += dt;` and in this class, it would do `real relaxation_parameter = compute_relaxation_parameter_explicit(); this->current_time += relaxation_parameter * dt;` (overriding the function)

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+    // Intention is to point to physics (mimic structure in flow_solver_cases/periodic_turbulence.cpp for converting to solution for general nodes) 
+    // For now, only energy on collocated nodes is implemented.
+    
+    real inner_product = 0;
+
+    bool use_collocated_nodes = ODESolverBase<dim,real,MeshType>::all_parameters->use_collocated_nodes;
+    using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
+    PDEEnum pde_type = ODESolverBase<dim,real,MeshType>::all_parameters->pde_type;
+    bool use_inviscid_burgers = (pde_type == PDEEnum::burgers_inviscid);
```
jbrillon, line 87: could just do `bool use_inviscid_burgers = this->all_parameters->pde_type == PDEEnum::burgers_inviscid`

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -0,0 +1,109 @@
...
+        )
+{
+    // Intention is to point to physics (mimic structure in flow_solver_cases/periodic_turbulence.cpp for converting to solution for general nodes) 
+    // For now, only energy on collocated nodes is implemented.
+    
+    real inner_product = 0;
+
+    bool use_collocated_nodes = ODESolverBase<dim,real,MeshType>::all_parameters->use_collocated_nodes;
+    using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
+    PDEEnum pde_type = ODESolverBase<dim,real,MeshType>::all_parameters->pde_type;
```
jbrillon, line 86: use keyword `this->` instead of `ODESolverBase<dim,real,MeshType>`

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return gamma;
+}
+
+template <int dim, typename real, typename MeshType>
+real RRKExplicitODESolver<dim,real,MeshType>::compute_inner_product (
+        dealii::LinearAlgebra::distributed::Vector<double> stage_i,
+        dealii::LinearAlgebra::distributed::Vector<double> stage_j
+        )
+{
+    // Intention is to point to physics (mimic structure in flow_solver_cases/periodic_turbulence.cpp for converting to solution for general nodes)
```
jbrillon, line 79: ah ok I see

jbrillon: could you open an issue for this ?

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -0,0 +1,109 @@
...
+    PDEEnum pde_type = ODESolverBase<dim,real,MeshType>::all_parameters->pde_type;
+    bool use_inviscid_burgers = (pde_type == PDEEnum::burgers_inviscid);
+    
+    if (use_collocated_nodes && use_inviscid_burgers){
+        for (unsigned int i = 0; i < this->dg->solution.size(); ++i) {
+            inner_product += 1./(this->dg->global_inverse_mass_matrix.diag_element(i))
+                             * stage_i[i] * stage_j[i];
+                }
+    }else {
+        this->pcout << "RRK is only implemented for inviscid burgers and collocated nodes." << std::endl;
```
jbrillon, line 95: Might be more appropriate for this to done at the factory level; at least to signal that RRK can't be used for `dim!=1`; I'd also comment out the appropriate/unused instantiations

src/ode_solver/rrk_explicit_ode_solver.cpp
jbrillon: The RRK scheme shouldn't depend on the `pde_type`; rather it should use `dg->pde_physics_double` by dynamically casting the passed `dg` shared_pointer as `DGBaseState` `shared_ptr`; see `src/numerical_flux/numerical_flux_factory.cpp` function `create_euler_based_convective_numerical_flux` as an example of this; the casting is needed since `DGBaseState` holds the physics objects.

src/ode_solver/rrk_explicit_ode_solver.h
```diff
@@ -0,0 +1,59 @@
...
+    /* Same as ExplicitODESolver except modifying timestep size
+     * No implementation for pseudotime
+     * Currently only explicit formulation for relaxation parameter is implemented 
+     * (i.e. Ketcheson 2019 method but not yet Ranocha 2020)
+     */
+    void step_in_time(real dt, const bool pseudotime) override;
+
+    /// Relaxation Runge-Kutta parameter gamma^n
+    /* See:  Ketcheson 2019, "Relaxation Runge--Kutta methods: Conservation and stability for inner-product norms"
+     *       Ranocha 2020, "Relaxation Runge--Kutta Methods: Fully Discrete Explicit Entropy-Stable Schemes for the Compressible Euler and Navier--Stokes Equations"
```
jbrillon, line 36: is the plan to eventually do something similar to this Ranocha paper?

src/parameters/parameters_ode_solver.cpp
```diff
@@ -122,6 +133,7 @@ void ODESolverParam::parse_parameters (dealii::ParameterHandler &prm)
             if (runge_kutta_order_string == "1") runge_kutta_order = 1;
             if (runge_kutta_order_string == "3") runge_kutta_order = 3;
             if (runge_kutta_order_string == "4") runge_kutta_order = 4;
+
```
jbrillon, line 136:
```suggestion
```

src/testing/burgers_energy_conservation_rrk.cpp
jbrillon: Did you forget to remove this function ? You should just be doing
`std::shared_ptr<FlowSolver::Periodic1DUnsteady<dim,nstate>> flow_solver_case = std::dynamic_pointer_cast<FlowSolver::Periodic1DUnsteady<dim,nstate>>(flow_solver->flow_solver_case);`
 and then `return flow_solver_case->compute_energy_collocated(flow_solver->dg)`

src/testing/burgers_energy_conservation_rrk.cpp
```diff
@@ -0,0 +1,161 @@
...
+    double energy = 0.0;
+    for (unsigned int i = 0; i < dg->solution.size(); ++i)
+    {
+        energy += 1./(dg->global_inverse_mass_matrix.diag_element(i)) * dg->solution(i) * dg->solution(i);
+    }
+    return energy;
+}
+
+template <int dim, int nstate>
+int BurgersEnergyConservationRRK<dim,nstate>::run_flow_solver(
```
jbrillon, line 73: rename to `get_energy_and_compare_to_initial`? Since its doing more than just `flow_solver->run()`

src/testing/burgers_energy_conservation_rrk.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    double final_time = this->all_parameters->flow_solver_param.final_time;
+    double time_step_large = this->all_parameters->ode_solver_param.initial_time_step;
+    double time_step_small = time_step_large * 1E-2;
+
+    int n_steps = round(final_time/time_step_large);
+    if (n_steps * time_step_large != final_time){
+        pcout << "Error: final_time is not evenly divisible by initial_time_step!" << std::endl
+              << "Remainder is " << fmod(final_time, time_step_large)
+              << ". Modify parameters to run this test." << std::endl;
```
jbrillon, line 97: Instead of having to modify the parameters manually, could you create a parameters object thats a copy of  `this->all_parameters` called `adjusted_parameters`, adjust the final_time so that it does satisfy this criteria, and pass `adjusted_parameters` when creating the flow solver

src/testing/burgers_energy_conservation_rrk.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    int testfail = 0;
+    int failed_this_calculation = 0;
+    
+    // Get initial energy
+    pcout << "\n\n-------------------------------------------------------------" << std::endl;
+    pcout << "  Calculating initial energy..." << std::endl;
+    pcout << "-------------------------------------------------------------" << std::endl;
+    std::unique_ptr<FlowSolver::FlowSolver<dim,nstate>> flow_solver = FlowSolver::FlowSolverFactory<dim,nstate>::select_flow_case((this->all_parameters), parameter_handler);
+    const double energy_initial = compute_energy_collocated(flow_solver->dg); //no need to run as ode_solver is allocated during construction
```
jbrillon, line 109: same comment as before for making use of the FlowSolverCase object

src/testing/burgers_energy_conservation_rrk.cpp
jbrillon: and add a pcout saying that the time step was adjusted

tests/integration_tests_control_files/burgers_split_stability/CMakeLists.txt
```diff
@@ -5,3 +5,21 @@ add_test(
...
+# Initial condition is sinusoidal; boundaries are periodic
+# Energy is expected to be conserved in all cases except (large dt, no RRK)
+# Test will fail if energy is not conserved in the other three cases
+# ----------------------------------------
+
+configure_file(1D_burgers_energy_conservation_rrk.prm 1D_burgers_energy_conservation_rrk.prm COPYONLY)
+add_test(
+    NAME 1D_BURGERS_ENERGY_CONSERVATION_RRK
+  COMMAND mpirun -n 1 ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_1D -i ${CMAKE_CURRENT_BINARY_DIR}/1D_burgers_energy_conservation_rrk.prm
+  WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
```
jbrillon, line 24:
```suggestion
    COMMAND mpirun -n 1 ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_1D -i ${CMAKE_CURRENT_BINARY_DIR}/1D_burgers_energy_conservation_rrk.prm
    WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
```

tests/integration_tests_control_files/time_refinement_study/CMakeLists.txt
```diff
@@ -13,3 +15,20 @@ add_test(
...
+# L2 error calculated with respect to a reference solution with a very small timestep at the actual end time
+# RRK can be used by changing the ODE solver in the .prm file
+# Test will fail if the convergence order is not close to the expected order
+# ----------------------------------------
+# ----------------------------------------
+configure_file(time_refinement_study_burgers_explicit.prm time_refinement_study_burgers_explicit.prm COPYONLY)
+add_test(
+    NAME 1D_TIME_REFINEMENT_STUDY_BURGERS_EXPLICIT
+  COMMAND mpirun -np 1 ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_1D -i ${CMAKE_CURRENT_BINARY_DIR}/time_refinement_study_burgers_explicit.prm
+  WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
```
jbrillon, line 33:
```suggestion
    COMMAND mpirun -np 1 ${EXECUTABLE_OUTPUT_PATH}/PHiLiP_1D -i ${CMAKE_CURRENT_BINARY_DIR}/time_refinement_study_burgers_explicit.prm
    WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
```

tests/integration_tests_control_files/time_refinement_study/CMakeLists.txt
```diff
@@ -13,3 +15,20 @@ add_test(
   WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
```
jbrillon, line 15:
```suggestion
    WORKING_DIRECTORY ${TEST_OUTPUT_DIR}
```

src/flow_solver/flow_solver_factory.cpp
jbrillon, line 49: if the only difference between these is the `pde_type` you specify in the `.prm` file, consider colapsing these two flow types into `periodic_1d_unsteady`

src/testing/time_refinement_study_reference.h
```diff
@@ -0,0 +1,56 @@
...
+    
+    /// Run test
+    int run_test () const override;
+protected:
+    /// Number of times to solve for convergence summary
+    const int n_time_calculations;
+
+    /// Ratio to refine by
+    const double refine_ratio;
+
```
jbrillon, line 35:
```suggestion
```

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
cpethrick: Yep, I have looked there and was planning to use that as a template! It just necessitates adding an energy integration function to `burgers.h`, which I saw as out of scope of this PR.

src/ode_solver/rrk_explicit_ode_solver.cpp
cpethrick: I was considering doing something like that but was having some difficulty with the best way to structure it. 

The "stage construction" loop (beginning at line 23) is identical to that in `explicit_ode_solver.cpp`. However, the "stage assembly" loop (beginning at line 44) is modified by the relaxation parameter. And, as you pointed out, the time increment is also modified.

That would mean that the options are either to repeat code (as I currently have done) or have `step_in_time()` look something like:
`construct_rk_stages()` -> loop beginning at line 23, same for explicit and RRK
`assemble_rk_stages()` -> loop beginning at line 44, RRK modifies by calculating `relaxation_parameter` and scaling the update.
`update_current_time()` -> as you pointed out; RRK modifies the time increment.

When first writing the code, I thought it would be most straightforward/readable to repeat `step_in_time()`, but I could also restructure as I outlined above. Do you think the restructuring I've proposed is a good idea?

src/testing/burgers_energy_conservation_rrk.cpp
cpethrick: I think it's safer to keep it as-is, to leave it up to the user's discretion whether their chosen time-step size or their chosen final time is more important. If I made the change you're suggesting, some scenarios would have unintended behaviour - for example, if the user is trying to stop calculation just before a shock forms and has made a mistake in entering the timestep.

src/ode_solver/rrk_explicit_ode_solver.cpp
cpethrick: Fixed (via moving to factory) in latest commit.

src/ode_solver/rrk_explicit_ode_solver.cpp
cpethrick: You're right, that's better to do in `ode_solver_factory`. Changing in the next commit.

src/ode_solver/rrk_explicit_ode_solver.h
cpethrick: Yes, hopefully! The method of finding the relaxation parameter I've implemented only works when the conserved/dissipated variable is energy. If it's entropy or some other convex functional, then the relaxation parameter needs to be found implicitly, as proposed in the Ranocha paper.

src/testing/burgers_energy_conservation_rrk.cpp
cpethrick: Yep, that's a good suggestion. Changing in the next commit.

src/testing/burgers_energy_conservation_rrk.cpp
cpethrick: Thanks for pointing that out, changing in the next commit.

src/testing/burgers_energy_conservation_rrk.cpp
cpethrick: I had the same function in `burgers_energy_conservation_rrk` and `time_refinement_study_reference`. It looks like I only removed it from the latter file - sorry for that! It's fixed for both files now.

src/ode_solver/rrk_explicit_ode_solver.cpp
cpethrick: I'll definitely open an issue for moving all instances of energy calculation into `physics`, and thanks for the suggestions of other places with similar code. 

I still have some decision-making as far as the best way to structure this part. As we discussed last week, this part of the code uses the same calculation as energy, but with the residual instead of the solution, therefore it might be confusing to call some `calculate_energy()` function when it's actually calculating the inner product of residuals. However, it's might be necessary in order to make it work for multiple PDE types.

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
jbrillon: I think the actual integration operation should be done in `src/flow_solver_cases/periodic_cube_flow.cpp` and somehow allow all derived classes to pass what should be getting integrated -- so that we minimize code repetition. Then `burger.h` simply defines the energy calculation at a single point.

src/ode_solver/rrk_explicit_ode_solver.cpp
jbrillon: It looks like the unused instantiations are still there and there is still no safeguard for checking that `dim == 1`

src/ode_solver/rrk_explicit_ode_solver.h
jbrillon: Ok very cool

src/ode_solver/rrk_explicit_ode_solver.cpp
jbrillon: ah I see, maybe just reference this conversation and the other related ones when you open the issue

src/ode_solver/rrk_explicit_ode_solver.cpp
cpethrick: Yep, added a check for `dim=1` and commented out the extra instantiations.

src/flow_solver/flow_solver_factory.cpp
cpethrick: This is a good suggestion, I have changed them to both be `periodic_1D_unsteady` and any differences use the `pde_type` as an indicator.

dougshidong:
Following up on this open issue. @jbrillon , what is missing for this to be approved?

jbrillon:
@dougshidong there are still some pending changes on this PR and I think @cpethrick is currently on vacation

jbrillon:
My mistake, I just noticed that the comments I left are all pending, meaning they haven't been visible to everyone else. I will publish these very soon

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
```diff
@@ -21,24 +11,74 @@ namespace FlowSolver {
...
 template <int dim, int nstate>
 void Periodic1DUnsteady<dim, nstate>::compute_unsteady_data_and_write_to_table(
        const unsigned int current_iteration,
         const double current_time,
-        const std::shared_ptr <DGBase<dim, double>>/* dg */,
-        const std::shared_ptr <dealii::TableHandler> /* unsteady_data_table */)
+        const std::shared_ptr <DGBase<dim, double>> dg ,
+        const std::shared_ptr <dealii::TableHandler> unsteady_data_table )
 {
     double dt = this->all_param.ode_solver_param.initial_time_step;
```
jbrillon, line 42:
```suggestion
    const double dt = this->all_param.ode_solver_param.initial_time_step;
```

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
```diff
@@ -21,24 +11,74 @@ namespace FlowSolver {
...
+        if ((current_iteration % output_solution_every_n_iterations) == 0){
+            this->pcout << "    Iter: " << current_iteration
+                        << "    Time: " << current_time
+                        << std::endl;
+        }
+        (void) dg;
+        (void) unsteady_data_table;
     }
+    else if ((pde_type == PDEEnum::burgers_inviscid)&&(this->all_param.use_collocated_nodes)){
+        double energy = this->compute_energy_collocated(dg);
```
jbrillon, line 58:
```suggestion
        const double energy = this->compute_energy_collocated(dg);
```

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
```diff
@@ -21,24 +11,74 @@ namespace FlowSolver {
...
+            this->pcout << "    Iter: " << current_iteration
+                        << "    Time: " << current_time
+                        << "    Energy: " << energy
+                        << std::endl;
+        }
+    
+        //detecting if the current run is calculating a reference solution 
+        int number_timesteps_ref = this->all_param.time_refinement_study_param.number_of_timesteps_for_reference_solution;
+        double final_time = this->all_param.flow_solver_param.final_time;
+        bool is_reference_solution = (dt < 2 * final_time/number_timesteps_ref);
```
jbrillon, line 70:
```suggestion
        const int number_timesteps_ref = this->all_param.time_refinement_study_param.number_of_timesteps_for_reference_solution;
        const double final_time = this->all_param.flow_solver_param.final_time;
        const bool is_reference_solution = (dt < 2 * final_time/number_timesteps_ref);
```

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -15,8 +16,20 @@ std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,Mesh
...
-    if(ode_solver_type == ODEEnum::implicit_solver) return std::make_shared<ImplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::explicit_solver)        return std::make_shared<ExplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::implicit_solver)        return std::make_shared<ImplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::rrk_explicit_solver)  {
+        //RRK is only implemented for Burgers on collocated nodes    
+        bool use_collocated_nodes = dg_input->all_parameters->use_collocated_nodes;
+        using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
+        PDEEnum pde_type = dg_input->all_parameters->pde_type;
+        bool use_inviscid_burgers = (pde_type == PDEEnum::burgers_inviscid);
+        if (use_collocated_nodes && use_inviscid_burgers && dim == 1)  return std::make_shared<RRKExplicitODESolver<dim,real,MeshType>>(dg_input);
```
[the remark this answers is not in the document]
jbrillon: applies to `create_ODESolver_manual()` as well

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -15,8 +16,20 @@ std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,Mesh
...
-    if(ode_solver_type == ODEEnum::explicit_solver) return std::make_shared<ExplicitODESolver<dim,real,MeshType>>(dg_input);
-    if(ode_solver_type == ODEEnum::implicit_solver) return std::make_shared<ImplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::explicit_solver)        return std::make_shared<ExplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::implicit_solver)        return std::make_shared<ImplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::rrk_explicit_solver)  {
+        //RRK is only implemented for Burgers on collocated nodes    
+        bool use_collocated_nodes = dg_input->all_parameters->use_collocated_nodes;
+        using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
+        PDEEnum pde_type = dg_input->all_parameters->pde_type;
+        bool use_inviscid_burgers = (pde_type == PDEEnum::burgers_inviscid);
```
jbrillon, line 26:
```suggestion
        const bool use_collocated_nodes = dg_input->all_parameters->use_collocated_nodes;
        using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
        const PDEEnum pde_type = dg_input->all_parameters->pde_type;
        const bool use_inviscid_burgers = (pde_type == PDEEnum::burgers_inviscid);
```

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -40,8 +53,20 @@ template <int dim, typename real, typename MeshType>
...
-    if(ode_solver_type == ODEEnum::explicit_solver) return std::make_shared<ExplicitODESolver<dim,real,MeshType>>(dg_input);
-    if(ode_solver_type == ODEEnum::implicit_solver) return std::make_shared<ImplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::explicit_solver)        return std::make_shared<ExplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::implicit_solver)        return std::make_shared<ImplicitODESolver<dim,real,MeshType>>(dg_input);
+    if(ode_solver_type == ODEEnum::rrk_explicit_solver){
+        //RRK is only implemented for Burgers on collocated nodes    
+        bool use_collocated_nodes = dg_input->all_parameters->use_collocated_nodes;
+        using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
+        PDEEnum pde_type = dg_input->all_parameters->pde_type;
+        bool use_inviscid_burgers = (pde_type == PDEEnum::burgers_inviscid);
```
jbrillon, line 63:
```suggestion
        const bool use_collocated_nodes = dg_input->all_parameters->use_collocated_nodes;
        using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
        const PDEEnum pde_type = dg_input->all_parameters->pde_type;
        const bool use_inviscid_burgers = (pde_type == PDEEnum::burgers_inviscid);
```

src/parameters/parameters_time_refinement_study.cpp
```diff
@@ -0,0 +1,42 @@
+#include "parameters_time_refinement_study.h"
+
+#include <string>
+
+namespace PHiLiP {
+
+namespace Parameters {
+
+// Flow Solver inputs
```
jbrillon, line 9:
```suggestion
```

src/parameters/parameters_time_refinement_study.h
```diff
@@ -0,0 +1,33 @@
...
+#define __PARAMETERS_TIME_REFINEMENT_STUDY_H__
+
+#include <deal.II/base/parameter_handler.h>
+#include "parameters/parameters.h"
+
+namespace PHiLiP {
+
+namespace Parameters {
+
+/// Parameters related to the flow solver
```
jbrillon, line 11:
```suggestion
/// Parameters related to time refinement studies
```

jbrillon (changes requested):
Great work! I've left some final comments, will approve once they have been addressed. Sorry I forgot to publish these earlier!

cpethrick:
Thanks for the additional comments @jbrillon! I'll resolve them today and tomorrow.

src/ode_solver/rrk_explicit_ode_solver.h
```diff
@@ -0,0 +1,53 @@
...
+protected:
+
+    /// Compute relaxation parameter explicitly (i.e. if energy is the entropy variable)
+    // See Ketcheson 2019, Eq. 2.4
+    real compute_relaxation_parameter_explicit();
+
+    void modify_time_step (real &dt);
+
+    /// Compute inner product according to the nodes being used
+    /* This is the same calculation as energy, but using the residual instead of solution
```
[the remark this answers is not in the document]
cpethrick: I'll check other header files to make sure that I haven't made mistakes elsewhere.

src/ode_solver/ode_solver_factory.cpp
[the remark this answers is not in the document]
cpethrick: I'll change the structure to use `if constexpr(dim==1)` in both functions in the next commit.

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+    numerator *= 2;
+    gamma = (denominator < 1E-8) ? 1 : numerator/denominator;
+    return gamma;
+}
+
+template <int dim, typename real, typename MeshType>
+real RRKExplicitODESolver<dim,real,MeshType>::compute_inner_product (
+        dealii::LinearAlgebra::distributed::Vector<double> stage_i,
```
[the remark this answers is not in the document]
cpethrick: That's a really good point! I will update in the next commit and take note for future contributions.

src/ode_solver/rrk_explicit_ode_solver.h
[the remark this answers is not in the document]
cpethrick: I believe this is the only affected file - fixed the other improperly formatted comment in the latest commit.

src/ode_solver/rrk_explicit_ode_solver.h
```diff
@@ -0,0 +1,54 @@
...
+    /// Relaxation Runge-Kutta parameter gamma^n
+    /** See:  Ketcheson 2019, "Relaxation Runge--Kutta methods: Conservation and stability for inner-product norms"
+     *       Ranocha 2020, "Relaxation Runge--Kutta Methods: Fully Discrete Explicit Entropy-Stable Schemes for the Compressible Euler and Navier--Stokes Equations"
+     */
+    real relaxation_parameter;
+
+protected:
+
+    /// Compute relaxation parameter explicitly (i.e. if energy is the entropy variable)
+    // See Ketcheson 2019, Eq. 2.4
```
jbrillon, line 35:
```suggestion
    /// See Ketcheson 2019, Eq. 2.4
```

jbrillon (approved):
LGTM, one small fix

PR #156, opened 2022-06-22
Goal-Oriented Adaptive Sampling and Addition of Eigen Library

donovan97:
Pull request for goal-oriented adaptive sampling for reduced-order models and the addition of the [Eigen library](https://gitlab.com/libeigen/eigen) to the code.

The pull request introduces significant changes to the reduced-order modeling. All of the old reduced-order modeling code is replaced as it did not work for cases run on multiple processors. Unused code and tests were also removed. 

Summary of main changes:

Changes to `Functional` class:
- The partial second derivatives of the class that are `dealii::TrilinosWrappers::SparseMatrix` are changed to pointers. This is because the `dealii::TrilinosWrappers::SparseMatrix` class has a `virtual` destructor to use pointers to the class, see https://www.dealii.org/current/doxygen/deal.II/classTrilinosWrappers_1_1SparseMatrix.html#adbe89a97bc7ef4bbc16e5e04b72bb19e
As a result of the `virtual` destructor, I was encountering errors due to the `dealii::TrilinosWrappers::SparseMatrix` within the `Functional` class when the `Functional` object in the `functionalFactory` method (`pod_adaptive_sampling.cpp`) went out of scope. The solution was to make the `SparseMatrices` pointers, and use a pointer to the Functional base class for the `functionalFactory` function in `pod_adaptive_sampling.cpp`. The destructor for the `Functional` class is also made virtual as this is supposed to be the case for a base class.
**Update July 11:** The `functionalFactory` method in `pod_adaptive_sampling.cpp` was removed and instead the `FunctionalFactory` class in `functional.cpp` is being used.

ODE solver changes:
- Galerkin and Petrov-Galerkin ODE solvers have been updated to work in parallel and to ensure solution consistency (reduced-order solution at snapshot point is exactly the full-order solution).

Test changes:
- `reduced_order.cpp` now tests for consistency of the reduced-order model. Tested on Burgers Rewienski and NACA0012 FlowSolver cases.
- `pod_adaptive_sampling.cpp` performs an adaptive snapshot sampling for the reduced-order model. Tested on Burgers Rewienski and NACA0012.
- `adaptive_sampling_testing.cpp` is used for validation testing and producing additional results, it does not actually test anything. Tests using this will be commented out before merging.

reduced_order directory changes:
- Removal of all files previously in this directory. 
- `pod_interface.h` provides an interface for the types of POD (offline and online POD)
- `pod_basis_offline.cpp` computes a POD basis from previously computed snapshots
- `pod_basis_online.cpp` computes a POD basis "online", during the adaptive sampling process
- `min_max_scaler.cpp` scales data between 0 and 1, used before interpolating
- `rbf_interpolation.cpp` implements a radial basis function interpolation
- `nearest_neighbors.cpp` implements k nearest neighbors for use with adaptive sampling 
- `reduced_order_solution.cpp` stores information about a reduced order solution that is required in future iterations of adaptive sampling
- `rom_test_location.cpp` computes adjoint-based error estimates for adaptive sampling
- `halton.cpp` computes a Halton sequence. I did not write this file, no need to review it.

Addition of Eigen library:

The pull request also adds the Eigen library as a submodule (see .gitmodules). Therefore, after merging the PR, everyone will have to update their git submodules to download the Eigen library in the submodules directory. This may need to be done with the command:
`git submodule update --init --recursive`

All required tests are passing and the addition of Eigen is not causing issues on Narval.


src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -36,39 +120,105 @@ void PODGalerkinODESolver<dim,real,MeshType>::step_in_time (real dt, const bool
...
-    this->reduced_lhs_tmp->mmult(*this->reduced_lhs, *pod->getPODBasis()); // reduced_lhs = reduced_lhs_tmp*pod_basis
-
-    solve_linear(
-            *this->reduced_lhs,
-            *this->reduced_rhs,
-            *this->reduced_solution_update,
-            this->ODESolverBase<dim,real,MeshType>::all_parameters->linear_solver_param);
-
-    pod->getPODBasis()->vmult(this->solution_update, *this->reduced_solution_update);
+    const Epetra_CrsMatrix epetra_system_matrix = this->dg->system_matrix.trilinos_matrix();
```
PranshulThakur, line 123: `Epetra_CrsMatrix` has a virtual destructor too. Not sure if it would result in the same issue that led to switching to pointers for partial second derivatives of functional.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -36,39 +120,105 @@ void PODGalerkinODESolver<dim,real,MeshType>::step_in_time (real dt, const bool
...
+    Epetra_Map system_matrix_rowmap = epetra_system_matrix.RowMap();
+    const Epetra_CrsMatrix epetra_pod_basis = pod->getPODBasis()->trilinos_matrix();
```
PranshulThakur, line 125: Since `epetra_pod_basis` and `epetra_system_matrix` are bring created and deleted in every iteration, will it be more efficient to have them as members of the class (without const)?

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -6,22 +6,106 @@ namespace ODE {
...
+        : ImplicitODESolver<dim,real,MeshType>(dg_input)
+        , pod(pod)
+{}
 
 template <int dim, typename real, typename MeshType>
-void PODGalerkinODESolver<dim,real,MeshType>::step_in_time (real dt, const bool /*pseudotime*/)
+int PODGalerkinODESolver<dim,real,MeshType>::steady_state ()
 {
+    this->pcout << " Performing steady state analysis... " << std::endl;
+    //allocate_ode_system ();
```
PranshulThakur, line 17:
```suggestion

```

PranshulThakur (commented):
Quite good PR to add goal-oriented adaptive sampling to the code. I have added a few comments based on the files reviewed so far.

src/ode_solver/ode_solver_base.cpp
```diff
@@ -208,14 +208,13 @@ int ODESolverBase<dim,real,MeshType>::steady_state ()
...
     if (ode_param.output_solution_vector_modulo > 0) {
-        for (unsigned int i = 0; i < this->dg->solution.size(); ++i) {
-            solutions_table.add_value(
-                    "Steady-state solution:",
-                    this->dg->solution[i]);
-        }
-        solutions_table.set_precision("Steady-state solution:", 16);
         std::ofstream out_file(ode_param.solutions_table_filename + ".txt");
-        solutions_table.write_text(out_file);
+        dealii::LinearAlgebra::ReadWriteVector<double> write_dg_solution(this->dg->solution.size());
```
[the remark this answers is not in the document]
donovan97: This is voluntary, I wanted to avoid having a header. I also currently only want the solution from the last run. `advance_solution_time()' is using a different and older code that outputs a header. I don't use this anymore so I could remove it if nobody else is using it?

src/ode_solver/pod_galerkin_ode_solver.cpp
donovan97: It should be fine. The issue with the second partial derivatives of Functional seems to happen because they are member variables.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -6,22 +6,106 @@ namespace ODE {
...
+
+    this->pcout << " Evaluating right-hand side and setting system_matrix to Jacobian before starting iterations... " << std::endl;
     const bool compute_dRdW = true;
     this->dg->assemble_residual(compute_dRdW);
-    this->current_time += dt;
-    // Solve (M/dt - dRdW) dw = R
-    // w = w + dw
 
-    this->dg->system_matrix *= -1.0;
+    const Epetra_CrsMatrix epetra_system_matrix = this->dg->system_matrix.trilinos_matrix();
```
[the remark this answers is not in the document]
donovan97: This is not a bad idea, but the epetra_system_matrix and epetra_pod_basis are also used for other computations in step_in_time, so I need to have them there anyway. Also, the line `epetra_petrov_galerkin_basis.Multiply(true, epetra_right_hand_side, epetra_reduced_rhs);` means that a function would either need to take epetra_right_hand_side and epetra_reduced_rhs as arguments, or I would need to make those two class members, both of which are kind of excessive for a function that would only contain 3 lines.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -36,39 +120,105 @@ void PODGalerkinODESolver<dim,real,MeshType>::step_in_time (real dt, const bool
...
+        this->dg->solution += reference_solution;
+        this->dg->assemble_residual();
+        epetra_pod_basis.Multiply(true, epetra_right_hand_side, epetra_reduced_rhs);
+        epetra_reduced_rhs.Norm2(&new_residual);
+        new_residual /= this->dg->right_hand_side.size();
+        this->pcout << " Step length " << step_length << " . Old residual: " << initial_residual << " New residual: " << new_residual << std::endl;
+    }
+    this->pcout << "Full-order residual norm: " << this->dg->get_residual_l2norm() << std::endl;
 
-    this->linesearch();
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -6,22 +6,106 @@ namespace ODE {
...
+                << std::endl
+                << " Initial absolute residual norm: " << this->initial_residual_norm
+                << std::endl;
+
+    this->residual_norm = 1;
+    this->residual_norm_decrease = 1;
+
+    // Initial Courant-Friedrichs-Lax number
+    const double initial_CFL = this->all_parameters->ode_solver_param.initial_time_step;
+    this->CFL_factor = 1.0;
```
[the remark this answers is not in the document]
PranshulThakur: Perfect!

src/ode_solver/pod_petrov_galerkin_ode_solver.cpp
```diff
@@ -1,74 +1,228 @@
...
+    epetra_reduced_rhs.Norm2(&this->initial_residual_norm);
+    this->initial_residual_norm /= this->dg->right_hand_side.size();
+
+    this->pcout << " ********************************************************** "
+          << std::endl
+          << " Initial absolute residual norm: " << this->initial_residual_norm
+          << std::endl;
+
+    this->residual_norm = 1;
+    this->residual_norm_decrease = 1;
```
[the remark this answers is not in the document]
donovan97: It's probably best to keep them separate for now. They are not necessarily in their final form and could change in the future, so I would rather wait once I know that they won't be changing anymore.

src/reduced_order/nearest_neighbors.cpp
```diff
@@ -0,0 +1,85 @@
...
+namespace ProperOrthogonalDecomposition {
+
+NearestNeighbors::NearestNeighbors()
+        : snapshot_params()
+        , scaler()
+        , snapshots()
+{}
+
+void NearestNeighbors::updateSnapshots(const MatrixXd &snapshot_parameters, dealii::LinearAlgebra::distributed::Vector<double> snapshot){
+    snapshot_params = snapshot_parameters;
```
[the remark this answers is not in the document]
donovan97: I am copying to snapshot_param to keep a copy of the current snapshot parameters in the NearestNeighbors class since it's used in other functions.
I would prefer not to call `updateSnapshots()` in the constructor because this class is used throughout the pod_adaptive_sampling.cpp test, and every time a snapshot is collected it is added using the updateSnapshots() function. However, if it gets called in the constructor then the NearestNeighbors object needs to be created after the first snapshot is collected, rather than before collecting any snapshots as is done now. That would involve making an exception for the first snapshot collected (not to call updateSnapshots() twice) which would not be very elegant.

src/reduced_order/min_max_scaler.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+MatrixXd MinMaxScaler::fit_transform(const MatrixXd &parameters){
+    min.resize(parameters.cols());
+    max.resize(parameters.cols());
+    MatrixXd parameters_scaled(parameters.rows(), parameters.cols());
+    for(int j = 0 ; j < parameters.cols() ; j++){
+        min(j) = parameters.col(j).minCoeff();
+        max(j) = parameters.col(j).maxCoeff();
+        for(int k = 0 ; k < parameters.rows() ; k++){
+            parameters_scaled(k, j) = (parameters(k, j) - min(j)) / (max(j) - min(j));
```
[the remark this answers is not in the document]
donovan97: I have added a check to warn the user if max(j) = min(j), but that would imply that there is no parameter range so that shouldn't really happen. I modeled the class after a similar Scipy class https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html 
It's sort of implied that you need to fit the scaler before you transform the data. So you can't really be calling transform without first calling fit_transform.

src/reduced_order/pod_basis_offline.cpp
```diff
@@ -0,0 +1,156 @@
...
+    std::sort(files_in_directory.begin(), files_in_directory.end()); //Sort files so that the order is the same as for the sensitivity basis
+
+    for (const auto & entry : files_in_directory){
+        if(std::string(entry.filename()).std::string::find("solution_snapshot") != std::string::npos){
+            pcout << "Processing " << entry << std::endl;
+            file_found = true;
+            std::ifstream myfile(entry);
+            if(!myfile)
+            {
+                pcout << "Error opening file." << std::endl;
```
PranshulThakur, line 41: Maybe add std::abort() if file doesn't open (or file_found = false).

src/reduced_order/pod_basis_offline.cpp
```diff
@@ -0,0 +1,156 @@
...
+                    if (field.empty()){ //due to whitespace
+                        continue;
+                    } else {
+                        cols++;
+                    }
+                }
+                rows++;
+            }
+
+            snapshotMatrix.conservativeResize(rows, snapshotMatrix.cols()+cols); //Subtract 1 from row because of header row
```
PranshulThakur, line 61: Should it be rows-1?

src/reduced_order/pod_basis_offline.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    referenceState.reinit(reference_state.size());
+    for(unsigned int i = 0 ; i < reference_state.size() ; i++){
+        referenceState(i) = reference_state(i);
+    }
+
+    snapshotMatrix = snapshotMatrix.colwise() - reference_state;
+
+    Eigen::BDCSVD<MatrixXd, Eigen::DecompositionOptions::ComputeThinU> svd(snapshotMatrix);
+    MatrixXd pod_basis = svd.matrixU();
```
PranshulThakur, line 106: That does the SVD decomposition? Eigen seems quite convenient!

src/reduced_order/pod_basis_offline.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    epetra_basis.FillComplete(domain_map, system_matrix_map);
+
+    basis->reinit(epetra_basis);
+
+    return file_found;
+}
+
+template <int dim>
+std::shared_ptr<dealii::TrilinosWrappers::SparseMatrix> OfflinePOD<dim>::getPODBasis() {
```
PranshulThakur, line 144: Were `getPODBasis()` and `getReferenceState()` written to make members `basis` and `referenceState` protected/private (they are currently public)?

src/reduced_order/pod_basis_offline.cpp
donovan97: Fixed in most recent commit.

src/reduced_order/pod_basis_offline.cpp
donovan97: No, the comment is from a previous version when I was using a dealii table to output the solution. I have since changed that to avoid having a header line. I have removed the comment, sorry about that.

src/reduced_order/pod_basis_offline.cpp
donovan97: Very convenient, I highly recommend it!

src/reduced_order/pod_basis_offline.cpp
donovan97: This is an example of polymorphism. I have two POD classes: OfflinePOD and OnlinePOD. Online POD takes snapshots on the fly and computes a POD basis for the adaptive sampling. Offline POD reads some previously computed snapshots stored on the hard drive and computes a POD basis. In both cases, the end use is to have a POD basis that I can use (and I also need a reference state to ensure consistency). Both POD classes implement the interface POD defined in pod_interface.h, which requires defining the functions getPODBasis() and getReferenceState(). The reduced-order ODE solvers only interface with POD classes using the functions defined in the interface POD. As a result, the reduced-order ODE solvers can use OfflinePOD or OnlinePOD, since both implement POD. Accessing data polymorphically like this can only be done with functions, not member variables.
It's also generally good practice to keep member variables private, like you said.

src/reduced_order/halton.cpp
```diff
@@ -0,0 +1,896 @@
...
+
+//****************************************************************************80
+//
+//  Purpose:
+//
+//    HALTON computes an element of a Halton sequence.
+//
+//  Licensing:
+//
+//    This code is distributed under the GNU LGPL license.
```
dougshidong, line 25: I'm gonna assume that you somehow can't import that code through some submodules?

src/testing/adaptive_sampling_testing.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+    else{
+        this->pcout << "Invalid flow case. You probably forgot to specify a flow case in the prm file." << std::endl;
+        std::abort();
+    }
+    return parameters;
+}
+
+template <int dim, int nstate>
+std::shared_ptr<Functional<dim,nstate,double>> AdaptiveSamplingTesting<dim, nstate>::functionalFactory(std::shared_ptr<DGBase<dim, double>> dg) const
```
dougshidong, line 134: To be moved later on. Can you open an issue about that.

src/testing/reduced_order.cpp
```diff
@@ -1,191 +1,94 @@
...
         return 0;
     }else{
         pcout << "Failed!";
         return -1;
     }
 }
+
+template <int dim, int nstate>
+std::shared_ptr<Functional<dim,nstate,double>> ReducedOrder<dim, nstate>::functionalFactory(std::shared_ptr<DGBase<dim, double>> dg) const
+{
```
dougshidong, line 65: Same factory? Put it somewhere common.

dougshidong (commented):
Aaah yeah, those 2nd derivatives being destructed when DG goes out of scope sounds like it would've been hard to debug if you had references to them. 

shared_ptr is fine, but we might want to be careful since those huge matrices can now continue on existing after DG is done. In the future, worth considering copying into a unique_ptr, or have a function in DG that allows another object to "steal" its matrices through a move. It's really just because those matrices are big and we have to be careful handling them and not allowing them to exist when not needed.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -6,22 +6,105 @@ namespace ODE {
...
+
+    this->pcout << " ********************************************************** "
+                << std::endl
+                << " ODESolver steady_state stopped at"
+                << std::endl
+                << " Nonlinear iteration: " << this->current_iteration
+                << " residual norm: " << this->residual_norm
+                << std::endl
+                << " ********************************************************** "
+                << std::endl;
```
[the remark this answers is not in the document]
donovan97: I would rather not for now. It seems like the reduced-order ODE solvers do not require CFL ramping to converge, but I am not fully certain, so I'm leaving steady_state as is for now. Once I'm sure that this is the case, I will likely have a parent class for the Galerkin and Petrov-Galerkin ODE solvers and steady_state will be a very small function.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -36,39 +119,109 @@ void PODGalerkinODESolver<dim,real,MeshType>::step_in_time (real dt, const bool
...
+    Epetra_Vector epetra_reduced_rhs(epetra_pod_basis.DomainMap());
+    epetra_pod_basis.Multiply(true, epetra_right_hand_side, epetra_reduced_rhs);
+    Epetra_CrsMatrix epetra_reduced_lhs(Epetra_DataAccess::View, epetra_pod_basis.DomainMap(), pod->getPODBasis()->n());
+    Epetra_CrsMatrix epetra_reduced_lhs_tmp(Epetra_DataAccess::View, epetra_pod_basis.RowMap(), pod->getPODBasis()->n());
+    EpetraExt::MatrixMatrix::Multiply(epetra_system_matrix, false, epetra_pod_basis, false, epetra_reduced_lhs_tmp, true);
+    EpetraExt::MatrixMatrix::Multiply(epetra_pod_basis, true, epetra_reduced_lhs_tmp, false, epetra_reduced_lhs);
+    Epetra_Vector epetra_reduced_solution_update(Epetra_DataAccess::View, epetra_reduced_lhs.DomainMap(), reduced_solution_update.begin());
+    Epetra_LinearProblem linearProblem(&epetra_reduced_lhs, &epetra_reduced_solution_update, &epetra_reduced_rhs);
+
+    Amesos_BaseSolver* Solver;
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/reduced_order/halton.cpp
donovan97: That's right, I found the code as a cpp file that I downloaded.

src/testing/adaptive_sampling_testing.cpp
donovan97: I just realized that there is a functional factory class in functional.cpp. I therefore removed this function and I'm now using the functional factory class. I added lift, drag and solution_integral (for Burgers Rewienski) to the factory, see most recent commit.

src/testing/reduced_order.cpp
donovan97: See other comment.

src/reduced_order/rom_test_location.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    gradient = rom_solution->gradient;
+
+    Parameters::LinearSolverParam linear_solver_param;
+    linear_solver_param.linear_solver_type = Parameters::LinearSolverParam::direct;
+    solve_linear(*rom_solution->system_matrix_transpose, gradient*=-1.0, adjoint, linear_solver_param);
+
+    //Compute dual weighted residual
+    fom_to_initial_rom_error = 0;
+    fom_to_initial_rom_error = -(adjoint * rom_solution->right_hand_side);
```
PranshulThakur, line 35: Since adjoint and gradient are distributed vectors, summing errors from all processors might be required (or maybe not because direct and Amesos solvers do it internally? I am not sure).

src/testing/adaptive_sampling_testing.cpp
```diff
@@ -0,0 +1,143 @@
...
+        data_table->set_precision("Parameter 2", 16);
+        data_table->set_precision("FOM func", 16);
+        data_table->set_precision("ROM func", 16);
+        data_table->set_precision("Func error", 16);
+        data_table->set_precision("State error", 16);
+
+        std::ofstream data_table_file("adaptation_data_table.txt");
+        data_table->write_text(data_table_file, dealii::TableHandler::TextOutputFormat::org_mode_table);
+    }
+    return 0;
```
PranshulThakur, line 107: What is this test checking for (and since this is the only return statement, does this test always pass)?

src/reduced_order/pod_basis_online.h
```diff
@@ -0,0 +1,81 @@
...
+#include <Epetra_Map.h>
+#include <eigen/Eigen/Dense>
+#include <eigen/Eigen/SVD>
+
+namespace PHiLiP {
+namespace ProperOrthogonalDecomposition {
+using Eigen::MatrixXd;
+using Eigen::VectorXd;
+
+/// Class for Online Proper Orthogonal Decomposition basis
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/reduced_order/rbf_interpolation.cpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+    else if(kernel == "linear"){
+        return r;
+    }
+    else{
+        return std::pow(r, 2) * std::log(r);
+    }
+}
+
+VectorXd RBFInterpolation::evaluate(const RowVectorXd& evaluate_coordinate) const {
```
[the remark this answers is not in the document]
donovan97: Great catch, fixed in most recent commit.

src/reduced_order/rom_test_location.cpp
donovan97: I have double checked and the complete error is output, not the error on a given processor, therefore it is ok.

src/testing/adaptive_sampling_testing.cpp
donovan97: It's a test used for validation testing and producing additional results, it does not actually test anything. I have commented out the tests in the CMakeLists for this test.

src/testing/reduced_order.cpp
```diff
@@ -1,191 +1,72 @@
...
-    double petrov_galerkin_error_norm_sum = 0;
+    std::shared_ptr<ProperOrthogonalDecomposition::OfflinePOD<dim>> pod_petrov_galerkin = std::make_shared<ProperOrthogonalDecomposition::OfflinePOD<dim>>(flow_solver_petrov_galerkin->dg);
+    flow_solver_petrov_galerkin->ode_solver =  PHiLiP::ODE::ODESolverFactory<dim, double>::create_ODESolver_manual(ode_solver_type, flow_solver_petrov_galerkin->dg, pod_petrov_galerkin);
+    flow_solver_petrov_galerkin->ode_solver->allocate_ode_system();
+    auto functional_petrov_galerkin = FunctionalFactory<dim,nstate,double>::create_Functional(all_parameters->functional_param, flow_solver_petrov_galerkin->dg);
 
-    unsigned int current_iteration = 0;
+    flow_solver_implicit->ode_solver->steady_state();
+    flow_solver_galerkin->ode_solver->steady_state();
+    flow_solver_petrov_galerkin->ode_solver->steady_state();
```
jbrillon, line 38: why not just call `flow_solver_petrov_galerkin->run();`? (minus creating that functional)

jbrillon: and set the `ode_solver_type` by passing in a modified parameter object to `create_FlowSolver` like you did in PR #134

src/testing/pod_adaptive_sampling.h
```diff
@@ -0,0 +1,127 @@
...
+#include "ROL_LineSearchStep.hpp"
+#include "ROL_StatusTest.hpp"
+#include "ROL_Stream.hpp"
+#include "ROL_Bounds.hpp"
+#include "functional/lift_drag.hpp"
+#include "functional/functional.h"
+#include "reduced_order/halton.h"
+#include "reduced_order/min_max_scaler.h"
+#include "tests.h"
+
```
jbrillon, line 38: by including `flow_solver.h`, some of these includes aren't needed

src/testing/reduced_order.h
```diff
@@ -4,20 +4,36 @@
...
+#include "functional/lift_drag.hpp"
+#include "parameters/all_parameters.h"
+#include <deal.II/numerics/solution_transfer.h>
+#include "flow_solver/flow_solver.h"
+#include "flow_solver/flow_solver_factory.h"
+#include <deal.II/base/numbers.h>
+#include "dg/dg_factory.hpp"
+#include "ode_solver/ode_solver_factory.h"
+#include <iostream>
+#include "functional/functional.h"
```
jbrillon, line 18: by including `flow_solver.h`, some of these includes aren't needed like dg and ode. Also, include only things that the .h file needs here, and the .cpp ones should be moved to the .cpp file; this might explain it better: https://stackoverflow.com/a/15420975

src/testing/pod_adaptive_sampling.h
jbrillon: Same comment as before where I mentioned https://stackoverflow.com/a/15420975

src/testing/pod_adaptive_sampling.cpp
```diff
@@ -0,0 +1,454 @@
...
+    this->pcout << "Solving FOM at " << parameter << std::endl;
+    Parameters::AllParameters params = reinitParams(parameter);
+
+    std::unique_ptr<FlowSolver::FlowSolver<dim,nstate>> flow_solver = FlowSolver::FlowSolverFactory<dim,nstate>::select_flow_case(&params, parameter_handler);
+
+    // Solve implicit solution
+    auto ode_solver_type = Parameters::ODESolverParam::ODESolverEnum::implicit_solver;
+    flow_solver->ode_solver =  PHiLiP::ODE::ODESolverFactory<dim, double>::create_ODESolver_manual(ode_solver_type, flow_solver->dg);
+    flow_solver->ode_solver->allocate_ode_system();
+    flow_solver->ode_solver->steady_state();
```
jbrillon, line 302: same comment as before about just calling `flow_solver-run()`

src/testing/pod_adaptive_sampling.cpp
```diff
@@ -0,0 +1,454 @@
...
+    Parameters::AllParameters params = reinitParams(parameter);
+
+    std::unique_ptr<FlowSolver::FlowSolver<dim,nstate>> flow_solver = FlowSolver::FlowSolverFactory<dim,nstate>::select_flow_case(&params, parameter_handler);
+
+    // Solve implicit solution
+    auto ode_solver_type = Parameters::ODESolverParam::ODESolverEnum::pod_petrov_galerkin_solver;
+    flow_solver->ode_solver =  PHiLiP::ODE::ODESolverFactory<dim, double>::create_ODESolver_manual(ode_solver_type, flow_solver->dg, current_pod);
+    //flow_solver->dg->solution = nearest_neighbors->nearestNeighborMidpointSolution(parameter);
+    flow_solver->ode_solver->allocate_ode_system();
+    flow_solver->ode_solver->steady_state();
```
jbrillon, line 320: same comment as before about just calling `flow_solver-run()`

src/testing/pod_adaptive_sampling.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+        snapshot_parameters.resize(5,2);
+        snapshot_parameters  << parameter1_range[0], parameter2_range[0],
+                                parameter1_range[0], parameter2_range[1],
+                                parameter1_range[1], parameter2_range[1],
+                                parameter1_range[1], parameter2_range[0],
+                                0.5*(parameter1_range[1] - parameter1_range[0])+parameter1_range[0], 0.5*(parameter2_range[1] - parameter2_range[0])+parameter2_range[0];
+
+        this->pcout << snapshot_parameters << std::endl;
+
```
jbrillon, line 380:
```suggestion
```

src/testing/adaptive_sampling_testing.h
```diff
@@ -0,0 +1,54 @@
...
+#include "dg/dg_factory.hpp"
+#include "ode_solver/ode_solver_factory.h"
+#include "flow_solver/flow_solver.h"
+#include "flow_solver/flow_solver_factory.h"
+#include "pod_adaptive_sampling.h"
+#include "dg/dg.h"
+#include "parameters/all_parameters.h"
+#include "functional/functional.h"
+#include <eigen/Eigen/Dense>
+#include "functional/lift_drag.hpp"
```
jbrillon, line 18: same comment as before about the includes

src/testing/adaptive_sampling_testing.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+        std::unique_ptr<FlowSolver::FlowSolver<dim,nstate>> flow_solver = FlowSolver::FlowSolverFactory<dim,nstate>::select_flow_case(&params, parameter_handler);
+        auto ode_solver_type = Parameters::ODESolverParam::ODESolverEnum::pod_petrov_galerkin_solver;
+        std::shared_ptr<ProperOrthogonalDecomposition::OfflinePOD<dim>> pod_standard = std::make_shared<ProperOrthogonalDecomposition::OfflinePOD<dim>>(flow_solver->dg);
+        flow_solver->ode_solver =  PHiLiP::ODE::ODESolverFactory<dim, double>::create_ODESolver_manual(ode_solver_type, flow_solver->dg, pod_standard);
+        flow_solver->ode_solver->allocate_ode_system();
+        auto functional = FunctionalFactory<dim,nstate,double>::create_Functional(params.functional_param, flow_solver->dg);
+
+        flow_solver_implicit->ode_solver->steady_state();
+        flow_solver->ode_solver->steady_state();
```
jbrillon, line 79: same comment as before about using `flow_solver->run()`

src/testing/CMakeLists.txt
```diff
@@ -19,12 +19,13 @@ set(TESTS_SOURCE
...
-    reduced_order.cpp
-    reduced_order_pod_adaptation.cpp
-    fd_state_sensitivity_wrt_parameter.cpp
     dual_weighted_residual_mesh_adaptation.cpp
     taylor_green_vortex_energy_check.cpp
     taylor_green_vortex_restart_check.cpp
+    pod_adaptive_sampling.cpp
+    adaptive_sampling_testing.cpp
+    reduced_order.cpp
+    taylor_green_vortex_restart_check.cpp
```
jbrillon, line 28: This is already included

src/reduced_order/rom_test_location.h
```diff
@@ -0,0 +1,79 @@
...
+#include "pod_interface.h"
+#include "reduced_order_solution.h"
+#include "linear_solver/linear_solver.h"
+#include <eigen/Eigen/Dense>
+#include <eigen/Eigen/SparseCore>
+#include <Epetra_Vector.h>
+#include <EpetraExt_MatrixMatrix.h>
+#include <Epetra_LinearProblem.h>
+#include "Amesos.h"
+#include "Amesos_BaseSolver.h"
```
jbrillon, line 24: same comment about the includes

jbrillon (commented):
Great work! I've left some comments/changes, we'd definitely want to eventually unify `steady_state()` so that its no longer virtual, or whatever makes sense for moving the contents of it over into flow_solver.cpp as we've discussed before since things like mesh adaptation should not be in ode_solver_base.cpp @PranshulThakur

jbrillon:
I added @cpethrick since she's working with the ODE solver

src/testing/CMakeLists.txt
jbrillon: (the tgv restart)

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -55,21 +55,21 @@ std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,Mesh
 template <int dim, typename real, typename MeshType>
 std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,MeshType>::create_ODESolver_manual(Parameters::ODESolverParam::ODESolverEnum ode_solver_type, std::shared_ptr< DGBase<dim,real,MeshType> > dg_input)
```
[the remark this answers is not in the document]
donovan97: I think I had discussed this with @dougshidong when I first wrote this code. It's not generally considered good practice to use nullptr in the way you're describing, and the use of nullptr should be avoided when possible. We had therefore decided on having two different functions, one for implicit and explicit, and another one for the reduced-order solvers that also take a POD basis.

src/testing/reduced_order.cpp
donovan97: I have changed it to call `flow_solver_implicit->run()`. However, the reduced-order solvers do not support polynomial ramping and do not yet seem to need this anyway, therefore they should only ever call steady_state, so I will leave those as is. 
For the `ode_solver_type`, I think it is just faster and prettier to swap the ODE object for the one that I need, rather than add a function with a bunch of if statements to match an enum type that I specified.

src/ode_solver/ode_solver_base.cpp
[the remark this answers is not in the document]
donovan97: Good points, I have renamed them as you suggested. I have also removed the solution output from `advance_solution_time()` as no one seems to be using it.

src/flow_solver/flow_solver.h
```diff
@@ -6,6 +6,7 @@
 #include "physics/initial_conditions/initial_condition.h"
 #include "physics/physics.h"
 #include "parameters/all_parameters.h"
+#include "reduced_order/pod_basis_offline.h"
```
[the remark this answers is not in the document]
donovan97: Fixed in most recent commit.

src/functional/functional.h
```diff
@@ -705,6 +705,55 @@ class FunctionalErrorNormLpBoundary : public Functional<dim,nstate,real,MeshType
...
+    SolutionIntegral(
+            std::shared_ptr<PHiLiP::DGBase<dim,real, MeshType>> dg_input,
+            std::shared_ptr<PHiLiP::Physics::PhysicsBase<dim,nstate,FadFadType>> _physics_fad_fad,
+            const bool uses_solution_values = true,
+            const bool uses_solution_gradient = false)
+            : PHiLiP::Functional<dim,nstate,real,MeshType>(dg_input,_physics_fad_fad,uses_solution_values,uses_solution_gradient)
+    {}
+
+    template <typename real2>
+    /// Templated volume integrand
```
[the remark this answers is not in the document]
donovan97: I think its usually before, so I fixed this in the most recent commit.

src/testing/pod_adaptive_sampling.h
donovan97: Fixed in most recent commit.

src/testing/reduced_order.h
donovan97: Fixed in most recent commit.

src/testing/pod_adaptive_sampling.cpp
donovan97: Fixed in most recent commit.

src/testing/adaptive_sampling_testing.h
donovan97: Fixed in most recent commit.

src/testing/CMakeLists.txt
donovan97: Fixed in most recent commit.

src/reduced_order/rom_test_location.h
donovan97: Fixed in most recent commit.

cpethrick (commented):
Great PR. The addition of the Eigen library seems very convenient, and it looks like you've put significant effort towards improving the reduced-order capabilities of PHiLIP. There's already been some great discussion already about the ODE Solver section, which I contributed to where I saw fit. I have a few fairly minor comments.

Does the addition of the Eigen library necessitate an update to the documentation/readme/installation files? If so, that should be added to your PR.

src/flow_solver/flow_solver.cpp
```diff
@@ -35,12 +35,20 @@ FlowSolver<dim, nstate>::FlowSolver(
 , final_time(flow_solver_param.final_time)
 , input_parameters_file_reference_copy_filename(flow_solver_param.restart_files_directory_name + std::string("/") + std::string("input_copy.prm"))
 , dg(DGFactory<dim,double>::create_discontinuous_galerkin(&all_param, poly_degree, poly_degree, grid_degree, flow_solver_case->generate_grid()))
-, ode_solver(ODE::ODESolverFactory<dim, double>::create_ODESolver(dg))
 {
     flow_solver_case->set_higher_order_grid(dg);
     dg->allocate_system();
+
+    if(flow_solver_param.reduced_order){
```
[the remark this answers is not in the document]
jbrillon: Btw are you always gonna be passing an object of type `ProperOrthogonalDecomposition::OfflinePOD` to ODE solver ? because we could just have a POD factory that takes parameters and returns a `nullptr` if there's no POD being done, and then the POD object would be held by the `FlowSolver` class like `dg` and `ode_solver`

src/reduced_order/pod_interface.h
```diff
@@ -0,0 +1,30 @@
...
+#include <deal.II/numerics/vector_tools.h>
+#include <deal.II/lac/la_parallel_vector.h>
+#include <deal.II/lac/trilinos_sparse_matrix.h>
+
+namespace PHiLiP {
+namespace ProperOrthogonalDecomposition {
+
+/// Interface for POD
+template <int dim>
+class POD
```
jbrillon, line 13: any reason this class isn't called `PODBase` ?

src/functional/functional.h
```diff
@@ -705,6 +705,55 @@ class FunctionalErrorNormLpBoundary : public Functional<dim,nstate,real,MeshType
...
+    /// Non-template functions to override the template classes
+    real evaluate_volume_integrand(
+            const PHiLiP::Physics::PhysicsBase<dim,nstate,real> &physics,
+            const dealii::Point<dim,real> &phys_coord,
+            const std::array<real,nstate> &soln_at_q,
+            const std::array<dealii::Tensor<1,dim,real>,nstate> &soln_grad_at_q) const override
+    {
+        return evaluate_volume_integrand<>(physics, phys_coord, soln_at_q, soln_grad_at_q);
+    }
+    /// Non-template functions to override the template classes
```
jbrillon, line 746:
```suggestion
    
    /// Non-template functions to override the template classes
```

src/testing/reduced_order.cpp
jbrillon: If they don't support polynomial_ramping then we would want to change `if(flow_solver_param.steady_state_polynomial_ramping){ ... }` to `if(flow_solver_param.steady_state_polynomial_ramping && 
     ((ode_solver_type != ODEEnum::pod_galerkin_solver) 
      && (ode_solver_type != ODEEnum::pod_galerkin_solver))){ ... }`

so that a user doesnt mistakenly try doing polynomial ramping with a POD-type ODE solver

jbrillon: then would you be able to call `->run` for the other two `flow_solver` objects

src/testing/pod_adaptive_sampling.h
jbrillon: it looks like its the exact same list in the .h and .cpp files, is that correct?

src/ode_solver/ode_solver_factory.cpp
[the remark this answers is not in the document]
donovan97: @cpethrick Done in recent commit.

src/ode_solver/pod_galerkin_ode_solver.h
```diff
@@ -28,27 +35,26 @@ class PODGalerkinODESolver: public ImplicitODESolver<dim, real, MeshType>
     /// Destructor
     ~PODGalerkinODESolver() {};
 
+    /// Evaluate steady state solution.
+    int steady_state () override;
```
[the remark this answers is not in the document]
donovan97: Good point, done in recent commit.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -36,39 +119,106 @@ void PODGalerkinODESolver<dim,real,MeshType>::step_in_time (real dt, const bool
...
+        epetra_reduced_rhs.Norm2(&new_residual);
+        new_residual /= this->dg->right_hand_side.size();
+        this->pcout << " Step length " << step_length << " . Old residual: " << initial_residual << " New residual: " << new_residual << std::endl;
+    }
+    this->pcout << "Full-order residual norm: " << this->dg->get_residual_l2norm() << std::endl;
 
-    pod->getPODBasis()->vmult(this->solution_update, *this->reduced_solution_update);
+    if(iline == maxline && new_residual > initial_residual){
+        this->pcout << "No decrease in residual. Maximum line search interations reached." << std::endl;
+    }
```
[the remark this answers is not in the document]
donovan97: Unfortunately not. The issue is for the Petrov-Galerkin ODE solver, I would need to remultiply the system matrix by the POD basis again in a separate line search function, which is quite expensive. By keeping it in the same function, I can avoid doing that. It could be done for the Galerkin solver, but as I've mentioned in other comments, I am likely going to create a parent class between them in the future, so no point in doing that right now.

src/ode_solver/pod_galerkin_ode_solver.cpp
```diff
@@ -36,39 +119,106 @@ void PODGalerkinODESolver<dim,real,MeshType>::step_in_time (real dt, const bool
...
-    this->dg->evaluate_mass_matrices(do_inverse_mass_matrix);
+    reference_solution = this->dg->solution;
+    reference_solution.import(pod->getReferenceState(), dealii::VectorOperation::values::insert);
+
+    dealii::LinearAlgebra::distributed::Vector<double> initial_condition(this->dg->solution);
+    initial_condition -= reference_solution;
+
+    const Epetra_CrsMatrix epetra_pod_basis = pod->getPODBasis()->trilinos_matrix();
+    reduced_solution.reinit(pod->getPODBasis()->n());
+    reduced_solution *= 0;
```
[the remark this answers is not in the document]
donovan97: I think I just wanted to be sure it was 0. I tested without `*=0` and it works fine, therefore you can safely rely on `reinit()`. I have removed the `*=0`.

src/reduced_order/pod_interface.h
donovan97: No reason, I changed it to PODBase in recent commit.

jbrillon: Should also rename the file accordingly if it makes sense

src/ode_solver/pod_petrov_galerkin_ode_solver.h
```diff
@@ -2,10 +2,17 @@
...
-#include "reduced_order/pod_basis.h"
+#include "reduced_order/pod_interface.h"
 #include <deal.II/lac/trilinos_sparsity_pattern.h>
+#include <EpetraExt_MatrixMatrix.h>
+#include <Epetra_Vector.h>
+#include <deal.II/lac/la_parallel_vector.h>
+#include <Epetra_Vector.h>
+#include <Epetra_LinearProblem.h>
+#include "Amesos.h"
+#include "Amesos_BaseSolver.h"
```
jbrillon, line 16: same comment as i left for the includes; try to apply this throughout the entire PR

src/testing/reduced_order.cpp
donovan97: I added the check so a user doesnt mistakenly try doing polynomial ramping with a POD ODE solver. However, I still can't call run for the other flow_solver objects. I would still need to create a function with some if statements to match the solver type to the enum. Otherwise, they will all run implicit since that is what's in the parameter file and what the variable `ode_solver_type` is set to.

src/reduced_order/pod_interface.h
donovan97: Done in recent commit.

donovan97:
> Great PR. The addition of the Eigen library seems very convenient, and it looks like you've put significant effort towards improving the reduced-order capabilities of PHiLIP. There's already been some great discussion already about the ODE Solver section, which I contributed to where I saw fit. I have a few fairly minor comments.
> 
> Does the addition of the Eigen library necessitate an update to the documentation/readme/installation files? If so, that should be added to your PR.

Since Eigen was added as a submodule, git should handle it automatically for a new installation, like the other existing submodules. For anyone who already has PHiLiP installed, you will likely need to update your git submodules using the command:
`git submodule update --init --recursive`

src/ode_solver/pod_galerkin_ode_solver.cpp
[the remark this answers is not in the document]
cpethrick: Okay, glad to hear that works as intended. I've seen both ways done in other parts of the code, so I wasn't sure if I hadn't noticed a bug!

src/ode_solver/pod_galerkin_ode_solver.cpp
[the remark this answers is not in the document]
cpethrick: Okay, sounds good.

src/flow_solver/flow_solver.cpp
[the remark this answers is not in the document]
cpethrick: I saw that you've changed the condition to use ODESolverEnum, looks good! Is this the only place where `flow_solver_param.reduced_order` is used? If so, it could be removed from the parameters list.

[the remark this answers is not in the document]
donovan97: Yeah that was the only use of it. I have removed it from the parameters list in the newest commit.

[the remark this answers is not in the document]
cpethrick: Okay, sounds great!

src/functional/functional.h
```diff
@@ -705,6 +705,56 @@ class FunctionalErrorNormLpBoundary : public Functional<dim,nstate,real,MeshType
...
+            const std::array<dealii::Tensor<1,dim,real2>,nstate> &soln_grad_at_q) const;
+
+    /// Non-template functions to override the template classes
+    real evaluate_volume_integrand(
+            const PHiLiP::Physics::PhysicsBase<dim,nstate,real> &physics,
+            const dealii::Point<dim,real> &phys_coord,
+            const std::array<real,nstate> &soln_at_q,
+            const std::array<dealii::Tensor<1,dim,real>,nstate> &soln_grad_at_q) const override
+    {
+        return evaluate_volume_integrand<>(physics, phys_coord, soln_at_q, soln_grad_at_q);
```
jbrillon, line 744:
```suggestion
        return evaluate_volume_integrand<real>(physics, phys_coord, soln_at_q, soln_grad_at_q);
```

src/functional/functional.h
```diff
@@ -705,6 +705,56 @@ class FunctionalErrorNormLpBoundary : public Functional<dim,nstate,real,MeshType
...
+    }
+    
+    /// Non-template functions to override the template classes
+    FadFadType evaluate_volume_integrand(
+            const PHiLiP::Physics::PhysicsBase<dim,nstate,FadFadType> &physics,
+            const dealii::Point<dim,FadFadType> &phys_coord,
+            const std::array<FadFadType,nstate> &soln_at_q,
+            const std::array<dealii::Tensor<1,dim,FadFadType>,nstate> &soln_grad_at_q) const override
+    {
+        return evaluate_volume_integrand<>(physics, phys_coord, soln_at_q, soln_grad_at_q);
```
jbrillon, line 754:
```suggestion
        return evaluate_volume_integrand<real>(physics, phys_coord, soln_at_q, soln_grad_at_q);
```

src/testing/reduced_order.cpp
jbrillon: You could just pass a modified copy of the parameters object to `create_FlowSolver()`

donovan97: I could, but its extra code for nothing if I can just call steady_state directly. Also, this won't be a problem anymore once we introduce the different flow solver types as we discussed.

src/functional/functional.h
donovan97: This actually doesn't compile. You'll notice that all other functionals in functional.h have empty brackets. I believe it will deduce the right argument. Let's leave it as it is.

src/functional/functional.h
donovan97: See other comment.

src/ode_solver/ode_solver_factory.cpp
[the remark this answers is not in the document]
donovan97: I have merged them into a single function, see newest commit.

jbrillon (approved):

PR #157, opened 2022-06-27
Generate job parameter scripts from bash

jbrillon:
PR to add the option of generating cluster job submission file: `job_parameters_file.sh` using bash instead of manually editing it. Useful for quickly setting up many different jobs on the cluster.


dougshidong:
Looks fine to me. One thing I'd change is the default number of cores being used or some kind of comment about a rule of thumb of number of cores per DoFs. Not sure what's a good rule of thumb with the current code status.

jbrillon:
@dougshidong I agree but we'd have to wait for PR #111 to go through and do a scaling study since the Operators class will likely change things quite a bit.

dougshidong (approved):

PR #169, opened 2022-08-15
Jacobian-free Newton-Krylov implicit solver for diagonally-implicit RK schemes

cpethrick:
PR for introducing implicit RK using Jacobian-free Newton-Krylov (JFNK) for the implicit solver. JFNK can be quite fast as no Jacobian matrix needs to be stored. Rather, the Jacobian-vector product which arises during Newton iterations is approximated on-the-fly using finite differences. The absence of a Jacobian implies that automatic differentiation is not required. In the scope of this PR, there will be no preconditioner used.

This PR has also improved the RK ODE solver to use better object-oriented programming practices.

**Changes related to JFNK:**
- Rename `ExplicitODESolver` to `RungeKuttaODESolver`, as both implicit and explicit RK are handled by the same function. 
- Add diagonally implicit RK (DIRK) to `RungeKuttaODESolver`
- Create a JFNK solver using deal.ii's GMRES solver. Currently this and all associated classes are in `src/ode_solver/JFNK_solver`. `JFNK_solver.<cpp/h>` holds `deal.ii`'s GMRES solver. `jacobian_vector_product.<cpp/h>` exists mainly to act as a matrix-like operator for the GMRES solver, and also computes the unsteady residual.
- Added tests to demonstrate implicit RK convergence order for 1D advection and burgers'.

**Other changes to clean up RungeKuttaODESolver:**
- Added parameters for JFNK to `LinearSolverParameters`, re-using GMRES parameters where possible.
- `RungeKuttaODESolver` and derived class `RRKExplicitODESolver` now assign `n_rk_stages` as a template argument for better performance. `create_RungeKuttaODESolver` has been added to `ODESolverFactory` has been modified to assign `n_rk_stages` based on parameters. 
- Changed from selecting an RK method by its order to named methods using the parameter `runge_kutta_method`. This allows multiple RK methods to have the same order, and streamlines the addition of methods.
- The assignment of the Butcher tableau has been moved to classes derived from `RKTableauBase`. An associated factory, `create_RKTableau`, has been made.


src/ode_solver/runge_kutta_methods/rk_tableau_base.h
```diff
@@ -0,0 +1,40 @@
...
+{
+public:
+    /// Default constructor that will set the constants.
+    RKTableauBase(); 
+
+    /// Destructor
+    ~RKTableauBase() {};
+
+    /// Returns Butcher tableau "a" coefficient at position [i][j]
+    virtual double a(int i, int j);
```
donovan97, line 26: Unless there is a reason I am unaware of, if you want to keep the code the way it is now, you could make these functions a and b pure virtual in which case you wouldn't need a constructor in rk_tableau_base (similar to src/reduced_order/pod_basis_base.h) and you also wouldn't need a cpp file. This would be an interface or an abstract class. If you do this, you need to define functions a and b in all classes that implement this interface, as you've currently done. This would make the most sense if the functions a and b were defined differently in every class that implements rk_tableau_base. 
However, in this case, the functions a and b are identical in each subclass and the logic is contained only in the constructor. Therefore, it probably makes more sense to factor out the functions a and b and bring them in rk_tableau_base. Then your subclasses would only consist of their constructor.

donovan97 (commented):
Very nice PR, great work!

src/ode_solver/runge_kutta_methods/rk_tableau_base.h
cpethrick: I think your proposal for restructuring the RK tableau classes (i.e., `a()` and `b()` in the base class) would help to reduce repeated code. However, if I did this, there would be risk of segmentation fault - if a user somehow chose `RKTableauBase` as their RK method, there will be a segmentation fault if code tries to access locations in a table that haven't been initialized. Of course, if a user uses the factory they should never be able to choose `RKTableauBase`!
Can you think of a good way to make this safe?

donovan97: Making the constructor of `RKTableauBase` protected should prevent anyone from initializing a `RKTableauBase`, only the subclasses could be initialized.

cpethrick: Okay, I will try that -- thanks!

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,121 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+            pcout << " with 4 stages..." << std::endl;
+            return std::make_shared<RungeKuttaODESolver<dim,real,4,MeshType>>(dg_input,rk_tableau);
+        }
+        else{
+            pcout << "but number of stages was unclear. Aborting..." << std::endl;
+            std::abort();
+            return nullptr;
+        }
+    }
+    if constexpr(dim==1){
```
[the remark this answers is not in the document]
cpethrick: This is a cleaner structure. Implemented in latest commit.

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,121 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+    dealii::ConditionalOStream pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)==0);
+    using RKMethodEnum = Parameters::ODESolverParam::RKMethodEnum;
+    const RKMethodEnum rk_method = dg_input->all_parameters->ode_solver_param.runge_kutta_method;
+
+    const int n_rk_stages = dg_input->all_parameters->ode_solver_param.n_rk_stages;
+
+    if (rk_method == RKMethodEnum::ssprk3_ex)   return std::make_shared<SSPRK3Explicit<dim, real, MeshType>> (n_rk_stages);
+    if (rk_method == RKMethodEnum::rk4_ex)      return std::make_shared<RK4Explicit<dim, real, MeshType>>    (n_rk_stages);
+    if (rk_method == RKMethodEnum::euler_ex)    return std::make_shared<EulerExplicit<dim, real, MeshType>>  (n_rk_stages);
+    if constexpr(dim==1){
```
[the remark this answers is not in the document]
cpethrick: According to [this](https://stackoverflow.com/a/55492127), combining the `if constexpr` with another condition defeats the purpose of using `constexpr`. Therefore, I'll keep this section as-is.

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -3,38 +3,47 @@
...
-        : ExplicitODESolver<dim,real,MeshType>(dg_input)
-        {}
+template <int dim, typename real, int n_rk_stages, typename MeshType>
+RRKExplicitODESolver<dim,real,n_rk_stages,MeshType>::RRKExplicitODESolver(std::shared_ptr< DGBase<dim, real, MeshType> > dg_input,
+            std::shared_ptr<RKTableauBase<dim,real,MeshType>> rk_tableau_input)
+        : RungeKuttaODESolver<dim,real,n_rk_stages,MeshType>(dg_input,rk_tableau_input)
+{
+    using RKMethodEnum = Parameters::ODESolverParam::RKMethodEnum;
+    const RKMethodEnum rk_method = this->ode_param.runge_kutta_method;
+    if (rk_method == RKMethodEnum::euler_ex){
```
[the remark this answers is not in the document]
cpethrick: You're right, that's more consistent. All other checks are within the factory. Changed in the latest commit.

donovan97 (approved):

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -1,39 +1,26 @@
...
-                use_collocated_nodes && use_inviscid_burgers){
-            return std::make_shared<RRKExplicitODESolver<dim,real,MeshType>>(dg_input);
-        }
-        else{
-            display_error_ode_solver_factory(ode_solver_type, false);
-            return nullptr;
-        }
-    }
+    if(ode_solver_type == ODEEnum::runge_kutta_solver)         return create_RungeKuttaODESolver(dg_input);
+    if(ode_solver_type == ODEEnum::rrk_explicit_solver)        return create_RungeKuttaODESolver(dg_input);
```
jbrillon, line 22: maybe combine these if statements so were only calling that RK factory once

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -57,23 +44,9 @@ template <int dim, typename real, typename MeshType>
...
```
jbrillon, line 48: same comment

src/ode_solver/ode_solver_factory.cpp
jbrillon: could introduce a bool `is_runge_kutta_type_solver`

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+        if (n_rk_stages == 3){
+            pcout << " with 3 stages..." << std::endl;
+            return std::make_shared<RungeKuttaODESolver<dim,real,3,MeshType>>(dg_input,rk_tableau);
+        }
+        if (n_rk_stages == 4){
+            pcout << " with 4 stages..." << std::endl;
+            return std::make_shared<RungeKuttaODESolver<dim,real,4,MeshType>>(dg_input,rk_tableau);
+        }
+        else{
+            pcout << "but number of stages was unclear. Aborting..." << std::endl;
```
jbrillon, line 132: maybe "invalid number of stages. Aborting..." is more appropriate

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,MeshType>::create_RungeKuttaODESolver(std::shared_ptr< DGBase<dim,real,MeshType> > dg_input)
+{
+    dealii::ConditionalOStream pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)==0);
+
+    std::shared_ptr<RKTableauBase<dim,real,MeshType>> rk_tableau = create_RKTableau(dg_input);
+   
+
+    const int n_rk_stages = dg_input->all_parameters->ode_solver_param.n_rk_stages;
+    using ODEEnum = Parameters::ODESolverParam::ODESolverEnum;
+    ODEEnum ode_solver_type = dg_input->all_parameters->ode_solver_param.ode_solver_type;
```
jbrillon, line 111:
```suggestion
    const ODEEnum ode_solver_type = dg_input->all_parameters->ode_solver_param.ode_solver_type;
```

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
     std::abort();
 }
 
+template <int dim, typename real, typename MeshType>
+std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,MeshType>::create_RungeKuttaODESolver(std::shared_ptr< DGBase<dim,real,MeshType> > dg_input)
+{
+    dealii::ConditionalOStream pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)==0);
+
+    std::shared_ptr<RKTableauBase<dim,real,MeshType>> rk_tableau = create_RKTableau(dg_input);
+
```
jbrillon, line 107:
```suggestion
```

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+    dealii::ConditionalOStream pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)==0);
+
+    std::shared_ptr<RKTableauBase<dim,real,MeshType>> rk_tableau = create_RKTableau(dg_input);
+   
+
+    const int n_rk_stages = dg_input->all_parameters->ode_solver_param.n_rk_stages;
+    using ODEEnum = Parameters::ODESolverParam::ODESolverEnum;
+    ODEEnum ode_solver_type = dg_input->all_parameters->ode_solver_param.ode_solver_type;
+    if (ode_solver_type == ODEEnum::runge_kutta_solver) {
+
```
jbrillon, line 113:
```suggestion
```

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+            return std::make_shared<RungeKuttaODESolver<dim,real,2,MeshType>>(dg_input,rk_tableau);
+        }
+        if (n_rk_stages == 3){
+            pcout << " with 3 stages..." << std::endl;
+            return std::make_shared<RungeKuttaODESolver<dim,real,3,MeshType>>(dg_input,rk_tableau);
+        }
+        if (n_rk_stages == 4){
+            pcout << " with 4 stages..." << std::endl;
+            return std::make_shared<RungeKuttaODESolver<dim,real,4,MeshType>>(dg_input,rk_tableau);
+        }
```
jbrillon, line 130: these can all be combined making use of the `n_rk_stages` variable within `if (n_rk_stages <= 4)`

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+            if (use_collocated_nodes && use_inviscid_burgers){
+                pcout << "Creating Relaxation Runge Kutta ODE Solver";
+                if (n_rk_stages == 1){
+                    pcout << " with 1 stage..." << std::endl;
+                    return std::make_shared<RRKExplicitODESolver<dim,real,1,MeshType>>(dg_input,rk_tableau);
+                }
+                if (n_rk_stages == 2){
+                    pcout << " with 2 stages..." << std::endl;
+                    return std::make_shared<RRKExplicitODESolver<dim,real,2,MeshType>>(dg_input,rk_tableau);
+                }
```
jbrillon, line 152: same comment as before (combine all if statements)

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+                    pcout << "but number of stages was unclear. Aborting..." << std::endl;
+                    std::abort();
+                    return nullptr;
+                }
+            } else {
+                pcout << "RRK has only been tested on collocated nodes with Burgers. Aborting..,"<<std::endl;
+                std::abort();
+            }
+        } else {
+            pcout << "RRK has only been tested on 1D calculations. Aborting..,"<<std::endl;
```
jbrillon, line 171:
```suggestion
            pcout << "\n Error: RRK has only been tested on 1D calculations. Aborting..."<<std::endl;
```

src/ode_solver/ode_solver_factory.cpp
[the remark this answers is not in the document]
jbrillon: typically only use `constexpr` with template parameters as the argument; https://docs.microsoft.com/en-us/cpp/cpp/constexpr-cpp?view=msvc-170

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
...
+            return nullptr;
+        } else return std::make_shared<EulerExplicit<dim, real, MeshType>>  (n_rk_stages);
+    }
+    if constexpr(dim==1){
+        //Only tested in 1D with Burgers and advection
+        using PDEEnum = Parameters::AllParameters::PartialDifferentialEquation;
+        const PDEEnum pde_type = dg_input->all_parameters->pde_type;
+        if ((pde_type==PDEEnum::burgers_inviscid) || (pde_type==PDEEnum::advection)){
+            if (rk_method == RKMethodEnum::euler_im)    return std::make_shared<EulerImplicit<dim, real, MeshType>>  (n_rk_stages);
+            if (rk_method == RKMethodEnum::dirk_2_im)   return std::make_shared<DIRK2Implicit<dim, real, MeshType>>  (n_rk_stages);
```
jbrillon, line 208: `implicit` instead of `im` in the enum members

src/ode_solver/runge_kutta_methods/rk_tableau_base.h
jbrillon:
```suggestion
    double a(const int i, const int j) const;
```

jbrillon: non-virtual since we're not overriding in the derived classes, and const everywhere since the arguments and class variables should not get modified. Also, consider renaming this to `get_a()` since its a basically a "getter". 

Thinking about it more, it may make more sense to have a pure virtual function `virtual set_tableau_a() = 0` (same for b), enforcing that each derived class only "sets" the tableau values. These would be private. If you template this class with `n_rk_stages` and `renit` the tables in the constructor of the base class, followed by `set_tableau_a()` and `set_tableau_b()` in the constructor of the base class, then all the derived classes could have empty constructors and only provide a definition for the `set_tableau` functions. Then use `constexpr` with `n_rk_stages` in the `create_RKTableau` factory

jbrillon (commented):
Leaving some initial comments, will review the rest once they've been addressed

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -125,6 +98,132 @@ void ODESolverFactory<dim,real,MeshType>::display_error_ode_solver_factory(Param
     std::abort();
 }
 
+template <int dim, typename real, typename MeshType>
+std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,MeshType>::create_RungeKuttaODESolver(std::shared_ptr< DGBase<dim,real,MeshType> > dg_input)
+{
+    dealii::ConditionalOStream pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)==0);
+
+    std::shared_ptr<RKTableauBase<dim,real,MeshType>> rk_tableau = create_RKTableau(dg_input);
```
jbrillon, line 106: (Read the long comment I left before this one) I think the RK base class should call this in the constructor since it's templated with `n_rk_stages`, rather than it being here

src/ode_solver/ode_solver_factory.cpp
cpethrick: I have combined them in the most recent commit. I think that it is okay to have two conditions rather than a bool `is_runge_kutta_type_solver`, but I definitely will add a bool if I have to add any more RK-based ODE solvers in the future.

src/ode_solver/ode_solver_factory.cpp
cpethrick: Also fixed.

src/ode_solver/ode_solver_factory.cpp
cpethrick: Good catch. I've added `const` in a few other places, too.

src/ode_solver/ode_solver_factory.cpp
cpethrick: Unfortunately, it seems like the hard-coded structure I've used is necessary because `n_rk_stages` is not known at compile-time. I've restructured the `pcout` statements to make the structure a little cleaner, and added a comment justifying the hard-coding.

src/ode_solver/ode_solver_factory.cpp
cpethrick: I've implemented the same `pcout` restructuring.

src/ode_solver/ode_solver_factory.cpp
cpethrick: I've added `Error` to all error messages, and fixed the accidental comma.

src/ode_solver/ode_solver_factory.cpp
cpethrick: I chose `_ex` an `_im` for the enum members for convenience while changing parameter files. `parameters_ode_solver` has a [comment](https://github.com/dougshidong/PHiLiP/pull/169/files#diff-98dccea52bc8d9f227575d1e06bc06b192ce40bc493026b1250f9ba5110e2bddR92) explaining the suffixes. Do you think this is clear enough, or should I use the full words "explicit" and "implicit"?

jbrillon: Ok that looks fine.

src/ode_solver/runge_kutta_methods/rk_tableau_base.h
cpethrick: I have changed the structure of `RKTableau` and its derived classes to use the setter/getter functionality. I decided against templating n_rk_stages because there shouldn't be a significant performance difference, and it's a bit tidier to pass it as an argument to the constructor rather than as a template.

cpethrick:
I added the "c" part of the butcher tableau to all of the RK methods. This is unused in the current RK solver, but I have added it because PR #111 uses it for some unsteady manufactured solutions.

src/ode_solver/rrk_explicit_ode_solver.cpp
```diff
@@ -25,8 +25,8 @@ real RRKExplicitODESolver<dim,real,n_rk_stages,MeshType>::compute_relaxation_par
     for (int i = 0; i < n_rk_stages; ++i){
         for (int j = 0; j < n_rk_stages; ++j){
             real inner_product = compute_inner_product(this->rk_stage[i],this->rk_stage[j]);
-            numerator += this->butcher_tableau->b(i) *this-> butcher_tableau->a(i,j) * inner_product; 
-            denominator += this->butcher_tableau->b(i)*this->butcher_tableau->b(j) * inner_product;
+            numerator += this->butcher_tableau->get_b(i) *this-> butcher_tableau->get_a(i,j) * inner_product;
```
jbrillon, line 28: could be made more efficient if you define `const double b_i = this->butcher_tableau->get_b(i)` before  `j` loop; that way you fetch it only n_rk_stages times, instead of fetching it n_rk_stages*n_rk_stages times

src/ode_solver/runge_kutta_methods/rk_tableau_base.cpp
```diff
@@ -5,18 +5,21 @@ namespace PHiLiP {
 namespace ODE {
 
 template <int dim, typename real, typename MeshType> 
-RKTableauBase<dim,real, MeshType> :: RKTableauBase ()
+RKTableauBase<dim,real, MeshType> :: RKTableauBase (int n_rk_stages)
```
jbrillon, line 8: `const` argument

src/ode_solver/runge_kutta_methods/rk_tableau_base.h
```diff
@@ -38,6 +37,15 @@ class RKTableauBase
 
     /// Butcher tableau "b"
     dealii::Table<1,double> butcher_tableau_b;
+    
+    /// Setter for butcher tableau
+    virtual void set_tableau() = 0;
```
jbrillon, line 48: If it's definition is always the same, define it in the base class

src/ode_solver/runge_kutta_ode_solver.cpp
```diff
@@ -75,11 +75,11 @@ void RungeKuttaODESolver<dim,real,n_rk_stages,MeshType>::step_in_time (real dt,
     //assemble solution from stages
     for (int i = 0; i < n_rk_stages; ++i){
         if (pseudotime){
-            const double CFL = butcher_tableau->b(i) * dt;
+            const double CFL = butcher_tableau->get_b(i) * dt;
```
jbrillon, line 78: `this->butcher_tableau` instead of `butcher_tableau`

src/ode_solver/runge_kutta_ode_solver.h
```diff
@@ -0,0 +1,53 @@
...
+    std::shared_ptr<RKTableauBase<dim,real,MeshType>> butcher_tableau;
+
+    /// Implicit solver for diagonally-implicit RK methods, using Jacobian-free Newton-Krylov 
+    /** This is initialized for any RK method, but solution-sized vectors are 
+     *  only initialized if there is an implicit solve
+     */
+    JFNKSolver<dim,real,MeshType> solver;
+    
+    /// Storage for the derivative at each Runge-Kutta stage
+    std::vector<dealii::LinearAlgebra::distributed::Vector<double>> rk_stage;
```
jbrillon, line 44: Maybe this is more appropriate? 
```suggestion
    std::vector<dealii::LinearAlgebra::distributed::Vector<double>> rk_stage_rhs;
```

jbrillon: Since you mentioned "rk_stage[i] stores the RHS at a partial time-step (not solution u)"

jbrillon: Its also fine to leave as is

src/ode_solver/runge_kutta_methods/runge_kutta_methods.cpp
```diff
@@ -9,6 +9,8 @@ void SSPRK3Explicit<dim,real,MeshType> :: set_tableau()
 {
     set_a();
     set_b();
+    set_c();
+    this->pcout << "Assigned RK method: 3rd order SSP (explicit)" << std::endl;
```
jbrillon, line 13: If the definition of this function does get moved to the base class, you could have a `const std::string rk_method_string_input` as an input argument to the `RKTableauBase` constructor for initializing member: `const std::string rk_method_string`; then in `set_tableau()` you have `set_a(); set_b(); set_c(); this->pcout << "Assigned RK method: " << this->rk_method_string << std::endl;`

src/ode_solver/rrk_explicit_ode_solver.cpp
cpethrick: Good catch. I changed that. After doing a quick scan, I don't see any other places where `get_*` is called enough to merit storing it.

src/ode_solver/runge_kutta_methods/rk_tableau_base.cpp
cpethrick: Changed in the last commit.

src/ode_solver/runge_kutta_methods/rk_tableau_base.h
cpethrick: Changed such that `set_tableau()` is defined in the base class, according to our discussion. This change means that `RungeKuttaODESolver::allocate_ode_system()` now has to call `butcher_tableau->set_tableau()`.

src/ode_solver/runge_kutta_ode_solver.cpp
cpethrick: Thanks for pointing that out. Fixed in latest commit.

src/ode_solver/runge_kutta_methods/runge_kutta_methods.cpp
cpethrick: I've implemented this.

src/ode_solver/runge_kutta_methods/runge_kutta_methods.cpp
```diff
@@ -0,0 +1,154 @@
...
+    this->butcher_tableau_c.fill(butcher_tableau_c_values);
+}
+
+//##################################################################
+template <int dim, typename real, typename MeshType>
+void DIRK2Implicit<dim,real,MeshType> :: set_a()
+{
+    // Pareschi & Russo DIRK, x = 1 - sqrt(2)/2
+    // see: wikipedia.org/wiki/List_of_Runge%E2%80%93Kutta_methods#Diagonally_Implicit_Runge%E2%80%93Kutta_methods
+    const double x = 0.2928932188134525; //=1-sqrt(2)/2
```
jbrillon, line 100: How come this isn't computed?

src/ode_solver/runge_kutta_methods/runge_kutta_methods.cpp
```diff
@@ -0,0 +1,154 @@
...
+template <int dim, typename real, typename MeshType>
+void DIRK2Implicit<dim,real,MeshType> :: set_c()
+{
+    const double x = 0.2928932188134525; //=1-sqrt(2)/2
+    const double butcher_tableau_c_values[2] = {x, 1.0-x};
+    this->butcher_tableau_c.fill(butcher_tableau_c_values);
+}
+
+//##################################################################
+template class SSPRK3Explicit<PHILIP_DIM, double, dealii::Triangulation<PHILIP_DIM> >;
```
jbrillon, line 121: should this only be instantiated for 1D?

cpethrick:
In the development of this PR, I also tested implicit RK using the AD-based linear solver already implemented in `PHiLiP`. I thought it would be interesting to compare the two now that JFNK is functioning well. I modified `runge_kutta_ode_solver.cpp` (see attachment) such that I could compare AD to JFNK for the test `1D_TIME_REFINEMENT_STUDY_ADVECTION_IMPLICIT`. A summary of the comparison is as follows:

- I needed to decrease the size of the first timestep from 0.2 to 0.01 in order for AD to converge. So, for the test set-up, JFNK is more robust. As a caveat, the timestep for AD to converge could likely be increased by increasing the GMRES tolerance.
- JFNK and AD both converge at the theoretical convergence rate, shown in the attachments.
- JFNK has an L2 error about 6 times smaller than AD.
- JFNK takes more than 2x the time taken by AD. This is to be expected for a serial computation - JFNK uses many more operations than AD (which can be faster on GPU calculations). However, the efficiency hasn't been optimized for either method.

I've attached the modified `runge_kutta_ode_solver.cpp` which uses AD, the `.prm` file with reduced initial timestep, as well as the full output of `1D_TIME_REFINEMENT_STUDY_ADVECTION_IMPLICIT` using AD and JFNK.

[runge_kutta_ode_solver_ADversion.cpp.txt](https://github.com/dougshidong/PHiLiP/files/9618302/runge_kutta_ode_solver_ADversion.cpp.txt)
[time_refinement_study_advection_implicit.prm.txt](https://github.com/dougshidong/PHiLiP/files/9618304/time_refinement_study_advection_implicit.prm.txt)
[time_refinement_study_advection_implicit_AD_convergence.txt](https://github.com/dougshidong/PHiLiP/files/9618305/time_refinement_study_advection_implicit_AD_convergence.txt)
[time_refinement_study_advection_implicit_JFNK_convergence.txt](https://github.com/dougshidong/PHiLiP/files/9618306/time_refinement_study_advection_implicit_JFNK_convergence.txt)

dougshidong:
Nice analysis. 

As a side note, you can assemble the matrix-vector product using AD such that it becomes a JFNM. The real comparison is AD vs Fréchet FD.

As you mentioned both have opportunities for code optimization to speed them up. The real saving we're obtaining is the memory usage as we get to higher and higher orders.

cpethrick:
I was discussing with @jbrillon how to best set up this PR for compatibility with PR #111. We decided that it would be easiest to rename the files `runge_kutta_ode_solver.<cpp/h>` back to `explicit_ode_solver.<cpp/h> such that GIT recognizes it as the same file. The parameter names, etc. are unchanged from previous commits.  I will open an issue to change the names to `runge_kutta_ode_solver.<cpp/h>` once  #111 has been merged into master.

dougshidong:
What's left for this PR to be merged?

cpethrick:
@dougshidong It should be ready to merge soon (within a week, I would guess). @jbrillon was going to do a final review in the coming days. I have confirmed that `ctest` behaves the same as before, and the time taken for explicit timestepping has not changed.

src/ode_solver/ode_solver_factory.cpp
```diff
@@ -1,39 +1,28 @@
...
-        else{
-            display_error_ode_solver_factory(ode_solver_type, false);
-            return nullptr;
-        }
-    }
+    const ODEEnum ode_solver_type = dg_input->all_parameters->ode_solver_param.ode_solver_type;
+    if((ode_solver_type == ODEEnum::runge_kutta_solver)||(ode_solver_type == ODEEnum::rrk_explicit_solver))     
+        return create_RungeKuttaODESolver(dg_input);
+    if(ode_solver_type == ODEEnum::implicit_solver)         
+        return std::make_shared<ImplicitODESolver<dim,real,MeshType>>(dg_input);
```
jbrillon, line 25: Add `pcout << "Creating ODE Solver..." << std::endl;` before returning the ImplicitODESolver; since it was removed at the top of the function

src/ode_solver/JFNK_solver/jacobian_vector_product.h
```diff
@@ -0,0 +1,59 @@
+#ifndef __JACOBIAN_VECTOR_PRODUCT__
+#define __JACOBIAN_VECTOR_PRODUCT__
+
+#include "dg/dg.h"
+
+namespace PHiLiP {
+namespace ODE{
+
+//template things
+//UPDATE WITH #IF STUFF
```
jbrillon, line 10: add the stuff

src/ode_solver/JFNK_solver/jacobian_vector_product.h
```diff
@@ -0,0 +1,59 @@
...
+
+    void reinit_for_next_timestep(double dt_input,
+                double epsilon_input,
+                dealii::LinearAlgebra::distributed::Vector<double> &previous_step_solution_input);
+
+    void reinit_for_next_Newton_iter(dealii::LinearAlgebra::distributed::Vector<double> &current_solution_estimate_input);
+
+    // Application of matrix to vector src. Write result into dst.
+    void vmult (dealii::LinearAlgebra::distributed::Vector<double> &dst,
+                const dealii::LinearAlgebra::distributed::Vector<double> &src) const;
```
jbrillon, line 27: would it be possible to give more descriptive variable names than `dst` and `src` ?

src/ode_solver/JFNK_solver/jacobian_vector_product.h
```diff
@@ -0,0 +1,59 @@
...
+
+    ~JacobianVectorProduct() {};
+
+    void reinit_for_next_timestep(double dt_input,
+                double epsilon_input,
+                dealii::LinearAlgebra::distributed::Vector<double> &previous_step_solution_input);
+
+    void reinit_for_next_Newton_iter(dealii::LinearAlgebra::distributed::Vector<double> &current_solution_estimate_input);
+
+    // Application of matrix to vector src. Write result into dst.
```
jbrillon, line 25:
```suggestion
    /// Application of matrix to vector src. Write result into dst.
```

src/ode_solver/JFNK_solver/jacobian_vector_product.h
```diff
@@ -0,0 +1,59 @@
...
+
+    JacobianVectorProduct(std::shared_ptr< DGBase<dim, real, MeshType> > dg_input);
+
+    ~JacobianVectorProduct() {};
+
+    void reinit_for_next_timestep(double dt_input,
+                double epsilon_input,
+                dealii::LinearAlgebra::distributed::Vector<double> &previous_step_solution_input);
+
+    void reinit_for_next_Newton_iter(dealii::LinearAlgebra::distributed::Vector<double> &current_solution_estimate_input);
```
jbrillon, line 23: add doxygen comments

src/ode_solver/JFNK_solver/jacobian_vector_product.h
```diff
@@ -0,0 +1,59 @@
...
+protected:
+
+    /// pointer to dg
+    std::shared_ptr<DGBase<dim,real,MeshType>> dg;
+
+    /// timestep size for implicit Euler step
+    double dt;
+    
+    /// small number for finite difference
+    double epsilon;
```
jbrillon, line 41: rename to `perturbation` ?

src/ode_solver/JFNK_solver/jacobian_vector_product.cpp
```diff
@@ -0,0 +1,80 @@
...
+                double epsilon_input,
+                dealii::LinearAlgebra::distributed::Vector<double> &previous_step_solution_input)
+{
+    dt = dt_input;
+    epsilon = epsilon_input;
+    previous_step_solution = previous_step_solution_input;
+}
+
+template <int dim, typename real, typename MeshType>
+void JacobianVectorProduct<dim,real,MeshType>::compute_dg_residual(dealii::LinearAlgebra::distributed::Vector<double> &dst, dealii::LinearAlgebra::distributed::Vector<double> &w) const
```
jbrillon, line 29:
```suggestion
void JacobianVectorProduct<dim,real,MeshType>::compute_dg_residual(dealii::LinearAlgebra::distributed::Vector<double> &dst, const dealii::LinearAlgebra::distributed::Vector<double> &w) const
```

src/ode_solver/JFNK_solver/jacobian_vector_product.cpp
```diff
@@ -0,0 +1,80 @@
...
+void JacobianVectorProduct<dim,real,MeshType>::reinit_for_next_Newton_iter(dealii::LinearAlgebra::distributed::Vector<double> &current_solution_estimate_input)
+{
+    current_solution_estimate = std::make_unique<dealii::LinearAlgebra::distributed::Vector<double>>(current_solution_estimate_input);
+    current_solution_estimate_residual = compute_unsteady_residual(*current_solution_estimate);
+}
+
+template <int dim, typename real, typename MeshType>
+void JacobianVectorProduct<dim,real,MeshType>:: reinit_for_next_timestep(double dt_input,
+                double epsilon_input,
+                dealii::LinearAlgebra::distributed::Vector<double> &previous_step_solution_input)
```
jbrillon, line 21: `const` arguments

src/ode_solver/JFNK_solver/jacobian_vector_product.cpp
```diff
@@ -0,0 +1,80 @@
...
+namespace PHiLiP{
+namespace ODE{
+
+template <int dim, typename real, typename MeshType>
+JacobianVectorProduct<dim,real,MeshType>::JacobianVectorProduct(std::shared_ptr< DGBase<dim, real, MeshType> > dg_input)
+    : dg(dg_input)
+{}
+
+template <int dim, typename real, typename MeshType>
+void JacobianVectorProduct<dim,real,MeshType>::reinit_for_next_Newton_iter(dealii::LinearAlgebra::distributed::Vector<double> &current_solution_estimate_input)
```
jbrillon, line 12: `const` arguments

src/ode_solver/JFNK_solver/jacobian_vector_product.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    dg->solution = w;
+    dg->assemble_residual();
+    
+    dg->global_inverse_mass_matrix.vmult(dg->solution, dg->right_hand_side);//dg->solution = IMM * RHS
+    dst = dg->solution;
+}
+
+template <int dim, typename real, typename MeshType>
+dealii::LinearAlgebra::distributed::Vector<double> JacobianVectorProduct<dim,real,MeshType>::compute_unsteady_residual(dealii::LinearAlgebra::distributed::Vector<double> &w, bool do_negate) const
```
jbrillon, line 39: `const` arguments

jbrillon (commented):
Great work restructuring the explicit ODE solver; I've left some comments

src/parameters/parameters_linear_solver.cpp
```diff
@@ -61,6 +61,21 @@ void LinearSolverParam::declare_parameters (dealii::ParameterHandler &prm)
...
+
+        prm.enter_subsection("JFNK options");
+        {
+            prm.declare_entry("newton_residual", "1E-7",
+                              dealii::Patterns::Double(),
+                              "Convergence tolerance for Newton iterations");
+            prm.declare_entry("newton_max_iterations", "100",
+                              dealii::Patterns::Integer(),
+                              "Maximum number of Newton iterations");
+            prm.declare_entry("epsilon_jacobian", "1.490116119384765625e-8",
```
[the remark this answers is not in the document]
cpethrick: Okay, I'll do that.

src/ode_solver/JFNK_solver/jacobian_vector_product.h
cpethrick: Sorry, I forgot to remove these comments. I'll replace with proper documentation for the class's functionality.

src/ode_solver/JFNK_solver/jacobian_vector_product.h
cpethrick: I can expand `dst` to `destination` (and likewise for the other function using that naming). I will change `src` to `w` and modify the comment to make it more clear that the Jacobian-vector product is returned. I originally named to be consistent with `deal.ii` conventions, but in this context it's a little more clear to change to more verbose naming.

src/ode_solver/JFNK_solver/jacobian_vector_product.h
cpethrick: Done, thanks for pointing that out.

src/ode_solver/JFNK_solver/jacobian_vector_product.h
cpethrick: Sure, I've changed to `fd_perturbation` to be more descriptive.

src/ode_solver/JFNK_solver/jacobian_vector_product.cpp
cpethrick: Changed.

src/ode_solver/JFNK_solver/jacobian_vector_product.cpp
cpethrick: Changed.

src/ode_solver/JFNK_solver/jacobian_vector_product.cpp
cpethrick: Changed.

src/ode_solver/ode_solver_factory.cpp
cpethrick: That line was only at the top of `create_ODESolver(dg)`, not `create_ODESolver(dg,pod)` or `create_ODESolver_manual`, so I assumed it was left here from previous debugging, and removed it. The print statements from the RK solvers are intended to confirm the number of stages, rather than to confirm that the factory has been called. Is that okay? Otherwise I'll add print statements to all of the functions.

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -33,22 +37,50 @@ void ExplicitODESolver<dim,real,MeshType>::step_in_time (real dt, const bool pse
         }//dt * sum(a_ij * k_j)
         
         this->rk_stage[i].add(1.0,this->solution_update); //u_n + dt * sum(a_ij * k_j)
+       
+        //implicit solve for diagonal element
+        if (this->butcher_tableau->get_a(i,i) != 0){
```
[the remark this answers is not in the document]
cpethrick: That's a good idea and could be extended in the future for other performance optimizations. I've implemented it.

src/ode_solver/runge_kutta_methods/rk_tableau_base.h
```diff
@@ -0,0 +1,68 @@
+#ifndef __RK_TABLEAU_BASE__
+#define __RK_TABLEAU_BASE__
+
+#include "dg/dg.h"
```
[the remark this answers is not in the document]
cpethrick: I've changed it to use the minimum files to build - that is, only the triangulation files. Everything builds & runs as before.

src/ode_solver/runge_kutta_methods/runge_kutta_methods.h
```diff
@@ -0,0 +1,150 @@
+#ifndef __RUNGE_KUTTA_METHODS__
+#define __RUNGE_KUTTA_METHODS__
+
+#include "rk_tableau_base.h"
+#include "dg/dg.h"
```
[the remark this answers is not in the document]
cpethrick: I have removed `dg.h`. Since the triangulation files are included in `rk_tableau_base.h`, they are not needed to build.

src/parameters/parameters_ode_solver.cpp
```diff
@@ -126,15 +133,34 @@ void ODESolverParam::parse_parameters (dealii::ParameterHandler &prm)
...
+        if (rk_method_string == "euler_im"){
+            runge_kutta_method = RKMethodEnum::euler_im;
+            n_rk_stages  = 1;
+            rk_order = 1;
+        }
+        if (rk_method_string == "dirk_2_im"){
+            runge_kutta_method = RKMethodEnum::dirk_2_im;
+            n_rk_stages  = 2;
+            rk_order = 2;
+        }
```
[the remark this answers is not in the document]
cpethrick: Sure, I agree that this is nicer. I'll also change the other places in this file that use successive `if` statements.

src/ode_solver/ode_solver_factory.cpp
jbrillon: Yeah add it to all the functions, would be good. Similar to how we have a print statement for Creating DG somewhere; for large runs without enough memory, its helpful for seeing where the code fails

cpethrick: For sure, that's a good reason! I've just added `"Creating ODE Solver..." ` at the top of every `create_ODESolver` function. I think that's the simplest way & will require the least maintenance going forward - RK-type solvers now print two lines but I think that's fine.

jbrillon (approved):
Great work

PR #170, opened 2022-08-29
Dedicated boundary ID for slip wall

jbrillon:
PR to incorporate the addition of a dedicated boundary ID for the slip wall boundary condition. This is needed by @kit71717 for testing 3D aerodynamic flow simulations.


donovan97 (approved):
Looks good to me.

PR #181, opened 2022-12-07
Higher Order VTK Output based on the maximum polynomial degree

cpethrick:
This PR enables higher-order VTK output for use with ParaView.

Specifically, the function `DGBase::output_results_vtk()`  has been modified to select the number of subdivisions from the maximum of `grid_degree` and `poly_degree`, which allows higher-order results to be written even if the grid is straight. 

A parameter, `enable_higher_order_vtk_output`, is added to toggle whether higher-order results are written, which is false by default. In some cases, it may be sufficient to visualize only (bi/tri)linear interpolations, such as if the grid is fine. As well, not all vtk visualization programs support higher-order Lagrange elements. ParaView is able to (see [this deal.ii wiki page ](https://github.com/dealii/dealii/wiki/Notes-on-visualizing-high-order-output)). From my testing, VisIt cannot handle higher-order solutions, and is unable to interpret the vtk files. I'm unsure about other vtk programs.

Additionally, I have added an optional parameter to `DGBase::output_results_vtk()`, `current_time`. This can be used to add the simulation time to the visualization. If `current_time` isn't passed, the time is written as `0.0`, consistent with previous behaviour. On ParaView, select Filters->Annotate Global Data to show the current time.

To demonstrate the new behaviour, I've attached a visualization of `MPI_VISCOUS_TAYLOR_GREEN_VORTEX_ENERGY_CHECK_LONG` after 400 iterations, both with and without higher-order output.
TGV_withHigherOrder
TGV_noHigherOrder

### DIFF
--- a/src/dg/dg.cpp
+++ b/src/dg/dg.cpp
@@ -1749,7 +1749,7 @@ public:
 
 
 template <int dim, typename real, typename MeshType>
-void DGBase<dim,real,MeshType>::output_face_results_vtk (const unsigned int cycle)// const
+void DGBase<dim,real,MeshType>::output_face_results_vtk (const unsigned int cycle, const double current_time)// const
 {
 
     DataOutEulerFaces<dim, dealii::DoFHandler<dim>> data_out;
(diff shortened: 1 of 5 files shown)


jbrillon (approved):
LGTM

PranshulThakur (approved):
LGTM!

Issue #174, opened 2022-09-21
Rename explicit_ode_solver.cpp/h


cpethrick:
PR #169 changes the `ExplicitODESolver` to `RungeKuttaODESolver`, but the file names are still `explicit_ode_solver.<cpp/h>` despite this naming. This has been done to make PR #111 have fewer conflicts.

Once PR #111 has been merged, this temporary naming can be changed to `runge_kutta_ode_solver.<cpp/h>` to be consistent with the class names.

PR #195, opened 2023-02-13
Relaxation Runge Kutta -- Entropy implementation

cpethrick:
This PR further develops the RRK implementation in PHiLiP. The major changes are:
- Split RRK ode solver into Root-finding (new) and Algebraic (same as was previously implemented) classes. 
- RK (explicit and implicit) can use the matrix-free `apply_inverse_global_mass_matrix/apply_global_mass_matrix`.
- Add Heun's method (explicit RK) and third-order diagonally-implicit RK; updated 2nd order DIRK as the original one was a placeholder.
- RRK classes no longer are derived from RungeKuttaODESolver. Rather, they are derived from an EmptyRRKBase, and RungeKuttaODESolver has an EmptyRRKBase type object. It does not do anything unless the user selects an RRK type.
- Fixed issue #252 . As a result, the CPLUS scaling studies fail because p is outside the allowable range of 2 to 5. I have changed them to use cHu, which passes and still has the same intention as the original tests (that is, FR does not add run-time computational cost).
- Added tests to verify that RRK works as intended.

This PR will close issue #174 and issue #252.


src/ode_solver/ode_solver_factory.cpp
```diff
@@ -146,38 +147,61 @@ std::shared_ptr<ODESolverBase<dim,real,MeshType>> ODESolverFactory<dim,real,Mesh
...
+        const PDEEnum pde_type = dg_input->all_parameters->pde_type;
+        using NumFluxEnum = Parameters::AllParameters::TwoPointNumericalFlux;
+        const NumFluxEnum two_point_num_flux_type = dg_input->all_parameters->two_point_num_flux_type;
+        
+        enum NumEntropyEnum {energy, entropy};
+        NumEntropyEnum numerical_entropy_type;
+        std::string numerical_entropy_string;
+        if (pde_type == PDEEnum::burgers_inviscid){
+            numerical_entropy_type = NumEntropyEnum::energy;
+            numerical_entropy_string = "Energy";
```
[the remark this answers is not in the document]
cpethrick: This implementation differentiates between the energy (algebraic) RRK formulation of Ketcheson and the entropy (root-finding) RRK formulation of Ranocha. I'm using the PDE type as an indicator for this -- the block starting at line 171 chooses either an `EnergyRRKODESolver` or `EntropyRRKODESolver`.

Do you think that a different naming convention would be clearer here? I agree that it may be confusing to differentiate numerical entropy as "energy" or "entropy" here.

src/flow_solver/flow_solver_cases/periodic_entropy_tests.cpp
```diff
@@ -0,0 +1,290 @@
...
+        for (unsigned int iquad=0; iquad<n_quad_pts; ++iquad) {
+
+            std::array<double,nstate> soln_state;
+            // Extract solution and gradient in a way that the physics ca n use them.
+            for(int istate=0; istate<nstate; istate++){
+                soln_state[istate] = soln_at_q[istate][iquad];
+            }
+            const double density = soln_state[0];                                                                                                                                                              
+            const double pressure = euler_physics->compute_pressure(soln_state);
+            const double entropy = log(pressure) - 1.4* log(density);
```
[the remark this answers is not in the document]
cpethrick: In the updated version of the file, this is fixed. `compute_numerical_entropy_function` is called through `euler_physics`, so changes to the gas properties should be handled well.

src/flow_solver/flow_solver_cases/periodic_entropy_tests.cpp
```diff
@@ -0,0 +1,290 @@
...
+            const double quadrature_entropy = -density*entropy/0.4;
+            //integrand_numerical_entropy_function = this->euler_physics->compute_numerical_entropy_function(soln_state); //SOMEHOW NOT REACHING PHYSICS
+            integrand_numerical_entropy_function = quadrature_entropy;
+            integral_numerical_entropy_function += integrand_numerical_entropy_function * quad_weights[iquad] * metric_oper.det_Jac_vol[iquad];
```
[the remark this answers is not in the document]
cpethrick: This form integrates across the solution nodes rather than using an inner product.

From our discussions in the fall semester, using `num-entropy = v M u` was not generally true for Euler FR. We may be able to use if for cDG, but I'm not sure if it is proven or robust.

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -415,6 +451,9 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
     if(do_calculate_numerical_entropy){
         this->pcout << "    Num. Entropy: " << std::setprecision(16) << numerical_entropy;
     }
+    if(is_rrk){
+        this->pcout << "    Rel. Param.: " << std::setprecision(16) << relaxation_parameter;
```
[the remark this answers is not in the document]
cpethrick: Good point. I will expand it here and in periodic_entropy_tests.

src/ode_solver/relaxation_runge_kutta/energy_rrk_ode_solver.cpp
```diff
@@ -0,0 +1,67 @@
...
+template <int dim, typename real, int n_rk_stages, typename MeshType>
+real EnergyRRKODESolver<dim,real,n_rk_stages,MeshType>::compute_relaxation_parameter(real & /*dt*/) const
+{
+    double gamma = 1;
+    double denominator = 0;
+    double numerator = 0;
+    for (int i = 0; i < n_rk_stages; ++i){
+        const double b_i = this->butcher_tableau->get_b(i);
+        for (int j = 0; j < n_rk_stages; ++j){
+            real inner_product = compute_inner_product(this->rk_stage[i],this->rk_stage[j]);
```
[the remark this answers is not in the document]
cpethrick: Yes, I have added this to the .cpp file.

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
```diff
@@ -18,18 +18,19 @@ Periodic1DUnsteady<dim, nstate>::Periodic1DUnsteady(const PHiLiP::Parameters::Al
...
-    for (unsigned int i = 0; i < dg->solution.size(); ++i)
-    {
-        energy += 1./(dg->global_inverse_mass_matrix.diag_element(i)) * dg->solution(i) * dg->solution(i);
-    }
-    return energy;
+    //Calculating energy via matrix-vector product
+    dealii::LinearAlgebra::distributed::Vector<double> temp;
+    temp.reinit(dg->solution);
+    if(this->all_param.use_inverse_mass_on_the_fly){
+        dg->apply_global_mass_matrix(dg->solution, temp);
```
sivanadarajah, line 29: If we are running with FR, would this allow the option of computing the energy with a different norm?

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -65,7 +65,9 @@ void RungeKuttaODESolver<dim,real,n_rk_stages,MeshType>::step_in_time (real dt,
             this->rk_stage[i] = solver.current_solution_estimate;
 
         } // u_n + dt * sum(a_ij * k_j) <explicit> + dt * a_ii * u^(i) <implicit>
-            
+        
+        compute_stored_quantities(i);
```
sivanadarajah, line 69: Could we consider changing the name of this function to something more descriptive?

src/ode_solver/ode_solver_factory.cpp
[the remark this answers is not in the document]
sivanadarajah: Understood. Just to restate what you have expressed, your energy denotes an algebraic approach to computing the relaxation approach; while entropy is through a root-finding approach. What if we wanted to do kinetic_energy_preserving? I guess, how we compute the relaxation parameter would not matter since it depends on whether the resulting equation is linear in \gamma or not.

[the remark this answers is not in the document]
cpethrick: That's correct. Any general nonlinear numerical entropy variable would use the "entropy" formulation. I suppose that kinetic energy preserving would also need to use the root-finding formulation.

If energy <u,u> is the numerical entropy, then the algebraic approach is used.

src/flow_solver/flow_solver_cases/periodic_1D_unsteady.cpp
cpethrick: This function computes energy as `u^T M_FR u`. 

The `apply_global_mass_matrix` function applies the modified mass matrix (i.e., \tilde{M} = M + K). Therefore, energy is computed in the correct broken Sobolev norm. For Burgers, I observe conservation even in c != cDG cases.

I would have to look at other parts of the code to see how to compute energy in different norms.

src/ode_solver/explicit_ode_solver.cpp
cpethrick: For sure. For now I will rename this to `store_stage_solutions()` and make the description in `explicit_ode_solver.h` clearer.

src/ode_solver/ode_solver_factory.cpp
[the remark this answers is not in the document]
sivanadarajah: I think a different naming convention would be easier to follow but its a minor point.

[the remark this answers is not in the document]
cpethrick: I could rename to "AlgebraicRRK" and "RootRRK".

src/flow_solver/flow_solver_cases/periodic_entropy_tests.cpp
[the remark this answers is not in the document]
sivanadarajah: Sounds perfect.

cpethrick:
I'm pening this PR for review now. All of the new tests I've added pass, and I'm currently running ctest to ensure that I haven't broken anything else.

PranshulThakur (commented):
Great PR on RRK and it was quite easy to review the files, especially with the comments explaining the code at various places. I have added a few comments.

src/dg/dg_base.hpp
```diff
@@ -929,8 +932,29 @@ class DGBase
...
+     * Used in entropy-RRK ODE solver.
+     * This is stored in dg such that both flow solver case and ode solver can access it. 
+     * flow solver cases have no access to ode solver. */
+    double entropy_M_norm=0;
+
+    /// Relaxation parameter
+    /** Used in RRK ODE solver.
+     * This is stored in dg such that both flow solver case and ode solver can access it. 
+     * flow solver cases have no access to ode solver. */
+    double relaxation_parameter=1;
```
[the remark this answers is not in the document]
cpethrick: I think it's okay as-is -- to me, it makes sense that `relaxation_parameter` would be associated with `relaxation_runge_kutta`. 

Could another reviewer please weigh in on this? I can change it if there is consensus that it's confusing.

src/ode_solver/relaxation_runge_kutta/runge_kutta_store_entropy.cpp
```diff
@@ -0,0 +1,176 @@
...
+    const unsigned int nstate = dim + 2;
+    // Currently only implemented for constant p
+    const unsigned int poly_degree = dg->get_max_fe_degree();
+    if (poly_degree != dg->get_min_fe_degree()){
+        this->pcout << "Error: Entropy RRK is only implemented for uniform p. Aborting..." << std::endl;
+        std::abort();
+    }
+    
+    // Select Euler physics. ODESolverFactory has already ensured that we are using Euler orSelect Euler physics. ODESolverFactory has already ensured that we are using Euler or NS, which both use the same entropy variable computation.
+    PHiLiP::Parameters::AllParameters parameters_euler = *(dg->all_parameters);
```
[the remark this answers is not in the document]
cpethrick: Changed to `parameters_using_euler_pde_type`. I don't want to use `all_parameters` to avoid confusion considering we use that name elsewhere, and this parameters object is only used to declare a physics class!

src/ode_solver/relaxation_runge_kutta/root_finding_rrk_ode_solver.cpp
```diff
@@ -0,0 +1,426 @@
...
+                soln_at_q[istate] = soln_at_q_vect[istate][iquad];
+                for(int idim=0; idim<dim; idim++){
+                    soln_grad_at_q[istate][idim] = soln_grad_at_q_vect[istate][idim][iquad];
+                }
+            }
+            
+            //#####################################################################
+            // Compute integrated quantities here
+            //#####################################################################
+            const double quadrature_entropy = euler_physics->compute_numerical_entropy_function(soln_at_q);
```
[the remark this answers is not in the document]
cpethrick: Thanks for pointing that out. I had removed the gradient calculations in `periodic_turbulence`, but forgot to do so here as well. Fixed & checked that the relevant test passes. Hopefully this improves computational cost :)

src/ode_solver/relaxation_runge_kutta/root_finding_rrk_ode_solver.cpp
```diff
@@ -0,0 +1,426 @@
...
+            double root_at_gamma = compute_root_function(gamma_kp1, u_n, step_direction, num_entropy_n, entropy_change_est,dg);
+            if (root_at_gamma < 0) {
+                l_limit = gamma_kp1;
+                root_l_limit = root_at_gamma;
+            } else {
+                u_limit = gamma_kp1;
+                root_u_limit = root_at_gamma;
+            }
+            residual = abs(root_at_gamma);
+            residual = u_limit-l_limit;
```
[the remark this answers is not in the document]
cpethrick: Yep, that's true. Fixed.

src/ode_solver/relaxation_runge_kutta/empty_RRK_base.h
```diff
@@ -0,0 +1,56 @@
...
+public:
+    /// Default constructor that will set the constants.
+    explicit EmptyRRKBase(std::shared_ptr<RKTableauBase<dim,real,MeshType>> /*rk_tableau_input*/);
+
+    /// Calculate FR entropy adjustment
+    /** Empty here
+     */
+    virtual double compute_FR_entropy_contribution(const real /*dt*/,
+            std::shared_ptr<DGBase<dim,real,MeshType>>/* dg*/,
+            const std::vector<dealii::LinearAlgebra::distributed::Vector<double>> &/*rk_stage*/,
```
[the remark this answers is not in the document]
cpethrick: Good point.

As a side note, I've tried to use `real` templates to maintain generality, but I make no guarantee that the parts of the code I've touched will work for anything except `double`!

src/flow_solver/flow_solver_cases/periodic_turbulence.h
```diff
@@ -77,7 +77,10 @@ class PeriodicTurbulence : public PeriodicCubeFlow<dim,nstate>
             const double current_time) const;
 
     /// Calculate numerical entropy by matrix-vector product
-    double get_numerical_entropy(const std::shared_ptr <DGBase<dim, double>> dg) const;
+    double calculate_numerical_entropy(const std::shared_ptr <DGBase<dim, double>> dg) const;
+    
+    /// Retrieves cumulative_numerical_entropy_change_FRcorrected
+    double get_numerical_entropy(const std::shared_ptr <DGBase<dim, double>> /*dg*/) const;
```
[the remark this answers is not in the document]
cpethrick: Good point. I have this naming convention such that both `periodic_turbulence` and `periodic_1D_unsteady` have a `get_numerical_entropy`, which can be called from a test. It results in a somewhat confusing naming convention within the `periodic_turbulence` case, as you've pointed out!

I will fix it by renaming `calculate_numerical_entropy` to `compute_current_integrated_numerical_entropy`. Hopefully it's clear enough that the `get_numerical_entropy` function is just retrieving an already-computed quantity.

src/flow_solver/flow_solver_cases/periodic_cube_flow.cpp
```diff
@@ -77,6 +77,20 @@ void PeriodicCubeFlow<dim,nstate>::display_grid_parameters() const
     if constexpr(dim==1) this->pcout << "- - Domain length: " << this->domain_size << std::endl;
     if constexpr(dim==2) this->pcout << "- - Domain area: " << this->domain_size << std::endl;
     if constexpr(dim==3) this->pcout << "- - Domain volume: " << this->domain_size << std::endl;
+
+    std::string c_parameter_string;
+    using FREnum = Parameters::AllParameters::Flux_Reconstruction;
+    FREnum fr_type = this->all_param.flux_reconstruction_type;
+    if (fr_type == FREnum::cDG)         c_parameter_string = "cDG";
+    if (fr_type == FREnum::cSD)         c_parameter_string = "cSD";
```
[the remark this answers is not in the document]
cpethrick: Good suggestion, fixed.

src/ode_solver/relaxation_runge_kutta/root_finding_rrk_ode_solver.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+template <int dim, typename real, typename MeshType>
+real RootFindingRRKODESolver<dim,real,MeshType>::compute_relaxation_parameter(const real dt,
+            std::shared_ptr<DGBase<dim,real,MeshType>> dg,
+            const std::vector<dealii::LinearAlgebra::distributed::Vector<double>> &rk_stage,
+            const dealii::LinearAlgebra::distributed::Vector<double> &solution_update
+            ) 
+{
+    // Console output is based on linearsolverparam
+    const bool do_output = (dg->all_parameters->linear_solver_param.linear_solver_output == Parameters::OutputEnum::verbose);
```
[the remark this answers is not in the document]
cpethrick: You're  right, I've moved it to the ODE section.

PranshulThakur (approved):
LGTM!

src/dg/dg_base.hpp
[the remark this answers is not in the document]
dougshidong: Unless it's directly stored in a RRK class, relaxation_parameter applies to any solver. I see the comment saying it's stored in DG so that both flow solver and ode solver can access it. Is there a better way? Is RRK something that defines both time and space discretization? Otherwise, ideally, we would keep time-discretization parameters outside of DG.

src/dg/dg_base.cpp
```diff
@@ -2619,13 +2619,15 @@ void DGBase<dim,real,MeshType>::apply_inverse_global_mass_matrix(
...
         dealii::LinearAlgebra::distributed::Vector<double> &output_vector,
-        const bool use_auxiliary_eq)
+        const bool use_auxiliary_eq,
+        const bool use_M_norm)
 {
     using FR_enum = Parameters::AllParameters::Flux_Reconstruction;
     using FR_Aux_enum = Parameters::AllParameters::Flux_Reconstruction_Aux;
-    const FR_enum FR_Type = this->all_parameters->flux_reconstruction_type;
+    const FR_enum FR_cDG = FR_enum::cDG;
+    const FR_enum FR_Type = (use_M_norm) ? FR_cDG : this->all_parameters->flux_reconstruction_type;
```
jbrillon, line 2630: what is going on here?

src/dg/dg_base.hpp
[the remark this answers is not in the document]
jbrillon: you could replace `flow_solver_case->compute_unsteady_data_and_write_to_table(ode_solver->current_iteration, ode_solver->current_time, dg, unsteady_data_table);` in `flow_solver.cpp` by `flow_solver_case->compute_unsteady_data_and_write_to_table(ode_solver, dg, unsteady_data_table);` which would be defined as a new virtual function in `FlowSolverCaseBase` with the default definition as just `flow_solver_case->compute_unsteady_data_and_write_to_table(ode_solver->current_iteration, ode_solver->current_time, dg, unsteady_data_table);`. Then in your flow solver case, you can override the definition for the new virtual function that takes the ode_solver as an argument and do whatever you need

src/flow_solver/flow_solver_cases/periodic_cube_flow.cpp
```diff
@@ -77,6 +77,20 @@ void PeriodicCubeFlow<dim,nstate>::display_grid_parameters() const
...
+    if (fr_type == FREnum::cDG)              c_parameter_string = "cDG";
+    else if (fr_type == FREnum::cSD)         c_parameter_string = "cSD";
+    else if (fr_type == FREnum::cHU)         c_parameter_string = "cHU";
+    else if (fr_type == FREnum::cNegative)   c_parameter_string = "cNegative";
+    else if (fr_type == FREnum::cNegative2)  c_parameter_string = "cNegative2";
+    else if (fr_type == FREnum::cPlus)       c_parameter_string = "cPlus";
+    else if (fr_type == FREnum::c10Thousand) c_parameter_string = "c10Thousand";
+    else if (fr_type == FREnum::cHULumped)   c_parameter_string = "cHULumped";
+
+    this->pcout << "- Using c parameter: " << c_parameter_string << std::endl;
```
jbrillon, line 93: maybe move this to `FlowSolverCaseBase<dim,nstate>::display_flow_solver_setup` and only print this if using strong form + flux reconstruction

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -734,53 +743,82 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
         const std::shared_ptr <dealii::TableHandler> unsteady_data_table)
 {
     // Compute and update integrated quantities
-    this->compute_and_update_integrated_quantities(*dg);
-    // Get computed quantities
+    if (do_calculate_overintegrated_quantities) this->compute_and_update_integrated_quantities(*dg);
```
jbrillon, line 746: maybe rename to `do_compute_and_update_integrated_quantities`; but this could cause issues if its false and an CFL number is used for the time step since the max wavelength is updated in the function `compute_and_update_integrated_quantities(*dg);`

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -734,53 +743,82 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
...
+        if(is_rrk) this->add_value_to_data_table(relaxation_parameter, "relaxation_parameter",unsteady_data_table);
+        if (do_calculate_overintegrated_quantities) this->add_value_to_data_table(integrated_kinetic_energy,"kinetic_energy",unsteady_data_table);
+        if (do_calculate_overintegrated_quantities) this->add_value_to_data_table(integrated_enstrophy,"enstrophy",unsteady_data_table);
+        if(is_viscous_flow && do_calculate_overintegrated_quantities) this->add_value_to_data_table(vorticity_based_dissipation_rate,"eps_vorticity",unsteady_data_table);
+        if (do_calculate_overintegrated_quantities) this->add_value_to_data_table(pressure_dilatation_based_dissipation_rate,"eps_pressure",unsteady_data_table);
+        if(is_viscous_flow && do_calculate_overintegrated_quantities) this->add_value_to_data_table(strain_rate_tensor_based_dissipation_rate,"eps_strain",unsteady_data_table);
+        if(is_viscous_flow && do_calculate_overintegrated_quantities) this->add_value_to_data_table(deviatoric_strain_rate_tensor_based_dissipation_rate,"eps_dev_strain",unsteady_data_table);
```
jbrillon, line 785: i think grouping some of these if statements would make it easier to read

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -734,53 +743,82 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
...
+    }
+    if(is_viscous_flow && do_calculate_overintegrated_quantities) {
         this->pcout << "    eps_vorticity: " << vorticity_based_dissipation_rate
                     << "    eps_p+eps_strain: " << (pressure_dilatation_based_dissipation_rate + strain_rate_tensor_based_dissipation_rate);
     }
     if(do_calculate_numerical_entropy){
-        this->pcout << "    Num. Entropy: " << std::setprecision(16) << numerical_entropy;
+        this->pcout << "    Num. Entropy cumulative, FR corrected: " << std::setprecision(16) << this->cumulative_numerical_entropy_change_FRcorrected; 
+        this->pcout << "    Num. Entropy change, FR corrected: " << std::setprecision(16) << current_numerical_entropy_change_FRcorrected; 
+        this->pcout << "    \\gamma \\Delta t sum bi v^T M u: " << std::setprecision(16) <<  dg->entropy_M_norm;
```
jbrillon, line 805: rather than a latex math expression, can you put a descriptive name instead?

src/flow_solver/flow_solver_cases/periodic_turbulence.h
```diff
@@ -141,6 +145,21 @@ class PeriodicTurbulence : public PeriodicCubeFlow<dim,nstate>
...
+
+    /// Numerical entropy at previous timestep
+    double previous_numerical_entropy = 0;
+
+    /// Cumulative change in numerical entropy
+    double cumulative_numerical_entropy_change_FRcorrected = 0;
+
+    /// Numerical entropy at initial time
+    double initial_numerical_entropy_abs = 0;
+
```
jbrillon, line 162: i think that maybe a class derived from `PeriodicTurbulence` might be more appropriate since you're introducing all these numerical entropy related things and its separate from Turbulence, maybe something like `PeriodicTurbulenceEntropyTests`? It would also allow you to replace the if statements by polymorphism

src/ode_solver/relaxation_runge_kutta/algebraic_rrk_ode_solver.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "dg/dg_base.hpp"
+#include "rrk_ode_solver_base.h"
+#include "ode_solver/ode_solver_base.h"
+#include "ode_solver/runge_kutta_ode_solver.h"
+
+namespace PHiLiP {
+namespace ODE {
+
+/// Relaxation Runge-Kutta ODE solver, calculating the relaxation parameter as in Ketcheson 2019
```
jbrillon, line 12: the full reference would be helpful instead of author year

src/ode_solver/relaxation_runge_kutta/algebraic_rrk_ode_solver.h
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+    /// Default constructor that will set the constants.
+    explicit AlgebraicRRKODESolver(
+            std::shared_ptr<RKTableauBase<dim,real,MeshType>> rk_tableau_input);
+
+protected:
+
+    /// Compute relaxation parameter explicitly (i.e. if energy is the entropy variable)
+    /// See Ketcheson 2019, Eq. 2.4
```
jbrillon, line 30: same comment for the full reference

src/ode_solver/relaxation_runge_kutta/algebraic_rrk_ode_solver.h
```diff
@@ -0,0 +1,49 @@
...
+#include "rrk_ode_solver_base.h"
+#include "ode_solver/ode_solver_base.h"
+#include "ode_solver/runge_kutta_ode_solver.h"
+
+namespace PHiLiP {
+namespace ODE {
+
+/// Relaxation Runge-Kutta ODE solver, calculating the relaxation parameter as in Ketcheson 2019
+/** "Relaxation Runge-Kutta Methods: Conservation and Stability for Inner-Product Norms" 
+ */
```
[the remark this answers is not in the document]
jbrillon: github didn't like that for some reason

src/ode_solver/relaxation_runge_kutta/empty_RRK_base.h
```diff
@@ -0,0 +1,56 @@
...
+#define __EMPTY_RRK_BASE_H__
+
+#include "dg/dg_base.hpp"
+#include "ode_solver/runge_kutta_methods/rk_tableau_base.h"
+
+namespace PHiLiP {
+namespace ODE {
+
+/// Empty class stored by RK solvers which do not use RRK, and also do not need to calculate numerical entropy.
+/// Functions in this class are empty or return a "dummy" value.
```
jbrillon, line 11: can they not just hold a `nullptr` by default? then you could make this a purely virtual class

src/ode_solver/runge_kutta_methods/runge_kutta_methods.cpp
```diff
@@ -95,25 +117,56 @@ void EulerImplicit<dim,real,MeshType> :: set_c()
...
+}
+
+//##################################################################
+template <int dim, typename real, typename MeshType>
+void DIRK3Implicit<dim,real,MeshType> :: set_a()
+{
+    // three-stage, stiffly-accurate SDIRK, gamma = 0.43586652150845899941601945
+    // see "Diagonally Implicit Runge-Kutta Methods for Ordinary Differential Equations. A Review"
+    // Kennedy & Carpenter
+    // Sec. 5.1.3
```
jbrillon, line 152: publication year? applies to other references you've commented in this file

src/ode_solver/runge_kutta_methods/runge_kutta_methods.h
```diff
@@ -128,6 +152,37 @@ class DIRK2Implicit: public RKTableauBase <dim, real, MeshType>
...
+    DIRK3Implicit(const int n_rk_stages, const std::string rk_method_string_input) 
+        : RKTableauBase<dim,real,MeshType>(n_rk_stages, rk_method_string_input) { }
+
+protected:
+
+    const double gam = 0.435866521508458999416019; ///< Constant "gamma" given by Kennedy & Carpenter
+    const double alpha = 1 - 4*gam + 2*gam*gam; ///< Constant "alpha" given by Kennedy & Carpenter
+    const double beta = -1 + 6*gam - 9*gam*gam + 3*gam*gam*gam; ///< Constant "beta" given by Kennedy & Carpenter
+    const double b2 = -3*alpha*alpha / 4.0 / beta; ///< Constant "b2" given by Kennedy & Carpenter
+    const double c2 = (2 - 9*gam + 6*gam*gam) / 3.0 / alpha; ///< Constant "c2" given by Kennedy & Carpenter
```
jbrillon, line 174: state the full reference in a comment above these constants

src/ode_solver/runge_kutta_methods/runge_kutta_methods.h
```diff
@@ -79,6 +79,31 @@ class EulerExplicit: public RKTableauBase <dim, real, MeshType>
 };
 
 
+///  Heun's method (explicit trapezoid rule; SSP 2,2)
```
jbrillon, line 82: reference?

src/physics/euler.cpp
```diff
@@ -1166,7 +1166,7 @@ ::boundary_pressure_outflow (
         for (int istate=0; istate<nstate; ++istate) {
             soln_bc[istate] = soln_int[istate];
         }
-    } 
+    }
```
jbrillon, line 1169: avoid committing changes where its just a space or added empty line

src/ode_solver/runge_kutta_ode_solver.cpp
```diff
@@ -114,6 +118,12 @@ void RungeKuttaODESolver<dim,real,n_rk_stages,MeshType>::allocate_ode_system ()
     if(this->all_parameters->use_inverse_mass_on_the_fly == false) {
         this->pcout << " evaluating inverse mass matrix..." << std::flush;
         this->dg->evaluate_mass_matrices(true); // creates and stores global inverse mass matrix
+        //RRK needs both mass matrix and inverse mass matrix
+        using ODEEnum = Parameters::ODESolverParam::ODESolverEnum;
+        ODEEnum ode_type = this->ode_param.ode_solver_type;
+        if (ode_type == ODEEnum::rrk_explicit_solver){
+            this->dg->evaluate_mass_matrices(false); // creates and stores global mass matrix
```
jbrillon, line 125: why is `evaluate_mass_matrices` called twice?

cpethrick: Because using true or false toggles whether the mass matrix or inverse mass matrix is called. As indicated in the comments, passing `true` results in the inverse mass matrix being stored, and passing `false` results in the mass matrix being stored. They are stored in different variables so there's no issue with calling it twice.

src/dg/dg_base.hpp
```diff
@@ -264,18 +264,21 @@ class DGBase
...
+    *   use_M_norm flag allows the unmodified mass matrix to be used 
+    *   for FR, i.e., use M rather than M+K.
     */
     void apply_global_mass_matrix(
-        dealii::LinearAlgebra::distributed::Vector<double> &input_vector,
+        const dealii::LinearAlgebra::distributed::Vector<double> &input_vector,
         dealii::LinearAlgebra::distributed::Vector<double> &output_vector,
-        const bool use_auxiliary_eq = false);
+        const bool use_auxiliary_eq = false,
+        const bool use_M_norm = false);
```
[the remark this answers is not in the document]
cpethrick: The `use_M_norm` parameter controls whether the mass matrix which will be applied is a modified or unmodified mass matrix. I have added to the comment above the definition. I need this for the implementation of the root solver RRK method. 
Let me know if it is still unclear.

src/dg/dg_base.cpp
cpethrick: the mass matrix is assembled as a modified mass matrix, so M+K. If we want to force an application of the unmodified mass matrix M, we choose cDG (which is zero and results in K=0), rather than the C parameter which was chosen in the .prm file. I will add a comment, let me know if it should be further clarified.

src/flow_solver/flow_solver_cases/periodic_cube_flow.cpp
cpethrick: Sure, that makes sense. I'll only check `use_weak_form`, since printing cDG will tell the user that FR isn't being used. I'll also add a line that prints whether strong or weak is being used since that's already checked.

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
cpethrick: Actually I will just remove this. I added it for a specific test and it's probably not useful enough to commit to master.

src/physics/euler.cpp
cpethrick: Sure, I'll reset this to how it is in master.

src/ode_solver/runge_kutta_ode_solver.cpp
jbrillon: Ahh okay got it, thanks!

src/flow_solver/flow_solver_cases/periodic_turbulence.h
cpethrick: I would prefer to keep it in one class as I think it's more consistent to have the flow solver case chosen only by `flow_type`. If I were to make an additional class, I would break this pattern. 

As a compromise, I have reorganized the numerical entropy calculation into a separate function to keep the code a bit more tidy.

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
cpethrick: Per discussion, I have removed this parameter.

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
cpethrick: That's definitely an oversight on my part, I forgot that the maximum wave speed is updated in that loop.
Per discussion, I have removed this parameter.

src/ode_solver/runge_kutta_methods/runge_kutta_methods.h
cpethrick: added.

src/ode_solver/runge_kutta_methods/runge_kutta_methods.h
cpethrick: added.

src/ode_solver/relaxation_runge_kutta/algebraic_rrk_ode_solver.h
cpethrick: I think it's sufficient to put the full reference only at the top of the .h file. I have also added the full reference to the root-solving equivalent.

src/ode_solver/relaxation_runge_kutta/algebraic_rrk_ode_solver.h
cpethrick: added the full reference.

cpethrick:
I have responded to all comments. @jbrillon, could you please review/resolve them?

I've also merged the most recent master in. I am re-running ctest to make sure the RRK and limiter changes work well together.

src/dg/dg_base.hpp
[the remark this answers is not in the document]
cpethrick: @jbrillon I like that idea, as it will allow everything to be contained in the ODE solver classes. However, I will need to do more restructuring than I have time for today, and other than this change I've addressed all comments. Therefore, is it okay if I make this change in a future PR which would only address this change? I will make sure to do it soon.

@PranshulThakur @dougshidong I have appended `_RRK_solver` to the variables I've added to DG. They would hopefully be removed in the future PR mentioned above.

src/ode_solver/relaxation_runge_kutta/empty_RRK_base.h
jbrillon: Ahh I see, yeah this works

src/dg/dg_base.cpp
```diff
@@ -2619,13 +2619,17 @@ void DGBase<dim,real,MeshType>::apply_inverse_global_mass_matrix(
...
         dealii::LinearAlgebra::distributed::Vector<double> &output_vector,
-        const bool use_auxiliary_eq)
+        const bool use_auxiliary_eq,
+        const bool use_M_norm)
 {
     using FR_enum = Parameters::AllParameters::Flux_Reconstruction;
     using FR_Aux_enum = Parameters::AllParameters::Flux_Reconstruction_Aux;
-    const FR_enum FR_Type = this->all_parameters->flux_reconstruction_type;
+    const FR_enum FR_cDG = FR_enum::cDG;
+    // if using only the M norm, set c=0 throuhg the choice of cDG, which results in K=0
```
jbrillon, line 2630:
```suggestion
    // if using only the M norm, set c=0 through the choice of cDG, which results in K=0
```

src/dg/dg_base.cpp
jbrillon: got it! thanks

src/flow_solver/flow_solver_cases/periodic_cube_flow.cpp
```diff
@@ -77,6 +77,7 @@ void PeriodicCubeFlow<dim,nstate>::display_grid_parameters() const
     if constexpr(dim==1) this->pcout << "- - Domain length: " << this->domain_size << std::endl;
     if constexpr(dim==2) this->pcout << "- - Domain area: " << this->domain_size << std::endl;
     if constexpr(dim==3) this->pcout << "- - Domain volume: " << this->domain_size << std::endl;
+
```
jbrillon, line 80:
```suggestion
```

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -683,16 +717,20 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
         unsteady_data_table->write_text(unsteady_data_table_file);
     }
     // Print to console
+    
     this->pcout << "    Iter: " << current_iteration
-                << "    Time: " << current_time
-                << "    Energy: " << integrated_kinetic_energy
+                << "    Time: " << current_time;
+    this->pcout << "    Energy: " << integrated_kinetic_energy
```
jbrillon, line 723: I don't think these are needed

cpethrick: sure, removed

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -703,6 +741,9 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
         std::abort();
     }
 
+    previous_time = current_time;
+    dt_target = this->get_adaptive_time_step(dg);
```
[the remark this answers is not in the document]
cpethrick: Doesn't look like it. They used to be used for calculating the relaxation parameter but now that's done another way. I have removed them.

src/dg/dg_base.hpp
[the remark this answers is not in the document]
cpethrick: Okay, changed the naming per your suggestion.

src/dg/dg_base.hpp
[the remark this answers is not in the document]
PranshulThakur: Sounds good. Thanks for making the change. I will leave it up to @jbrillon or @dougshidong to resolve this comment.

cpethrick:
`ctest` passes as expected. I've attached the log. Note that the run time is quite a bit longer than usual as I ran with 4 cores and was running other things on my computer at the same time.
[ctestPR195.log](https://github.com/dougshidong/PHiLiP/files/14525524/ctestPR195.log)

src/dg/dg_base.hpp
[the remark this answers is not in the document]
jbrillon: If there's no rush for this PR then I'd say just address in this one when you have time

src/flow_solver/flow_solver_cases/flow_solver_case_base.cpp
```diff
@@ -98,6 +98,31 @@ void FlowSolverCaseBase<dim,nstate>::display_flow_solver_setup(std::shared_ptr<D
...
+        if (fr_type == FREnum::cDG)              c_parameter_string = "cDG";
+        else if (fr_type == FREnum::cSD)         c_parameter_string = "cSD";
+        else if (fr_type == FREnum::cHU)         c_parameter_string = "cHU";
+        else if (fr_type == FREnum::cNegative)   c_parameter_string = "cNegative";
+        else if (fr_type == FREnum::cNegative2)  c_parameter_string = "cNegative2";
+        else if (fr_type == FREnum::cPlus)       c_parameter_string = "cPlus";
+        else if (fr_type == FREnum::c10Thousand) c_parameter_string = "c10Thousand";
+        else if (fr_type == FREnum::cHULumped)   c_parameter_string = "cHULumped";
+
+        this->pcout << "- - Using c parameter: " << c_parameter_string << std::endl;
```
jbrillon, line 120: only necessary if `use_split_form == true`; maybe add another if statement to distinguish between classical strong DG and split-form/NSFR

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -645,7 +651,29 @@ double PeriodicTurbulence<dim, nstate>::get_numerical_entropy(
...
+{
+
+    const double current_numerical_entropy = this->compute_current_integrated_numerical_entropy(dg);
+
+    if (current_iteration==0) {
+        this->previous_numerical_entropy = current_numerical_entropy;
+        this->initial_numerical_entropy_abs = abs(current_numerical_entropy);
+    }
+
+    const double current_numerical_entropy_change_FRcorrected = (current_numerical_entropy - previous_numerical_entropy + dg->FR_entropy_contribution_RRK_solver)/initial_numerical_entropy_abs;
```
jbrillon, line 673:
```suggestion
    const double current_numerical_entropy_change_FRcorrected = (current_numerical_entropy - this->previous_numerical_entropy + dg->FR_entropy_contribution_RRK_solver)/this->initial_numerical_entropy_abs;
```

jbrillon: just to keep things consistent

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -657,21 +685,27 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
 {
     // Compute and update integrated quantities
     this->compute_and_update_integrated_quantities(*dg);
-    // Get computed quantities
+    // Get computed quantities.
```
jbrillon, line 688:
```suggestion
    // Get computed quantities
```

jbrillon: to minimize merge conflicts

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -544,7 +544,13 @@ double PeriodicTurbulence<dim, nstate>::get_strain_rate_tensor_based_dissipation
 }
 
 template<int dim, int nstate>
-double PeriodicTurbulence<dim, nstate>::get_numerical_entropy(
+double PeriodicTurbulence<dim, nstate>::get_numerical_entropy(const std::shared_ptr <DGBase<dim, double>> /*dg*/) const
```
jbrillon, line 547: if this is a getter, you can remove the argument here

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -657,21 +685,27 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
...
+
+    if (do_calculate_numerical_entropy){
+        this->update_numerical_entropy(current_iteration, dg);
+    }
 
     if(this->mpi_rank==0) {
         // Add values to data table
         this->add_value_to_data_table(current_time,"time",unsteady_data_table);
-        if(do_calculate_numerical_entropy) this->add_value_to_data_table(numerical_entropy,"numerical_entropy",unsteady_data_table);
+        if(do_calculate_numerical_entropy) this->add_value_to_data_table(this->cumulative_numerical_entropy_change_FRcorrected,"numerical_entropy_scaled_cumulative",unsteady_data_table);
```
jbrillon, line 707:
```suggestion
        if(do_calculate_numerical_entropy) this->add_value_to_data_table(this->get_numerical_entropy(),"numerical_entropy_scaled_cumulative",unsteady_data_table);
```

jbrillon: if you do remove the argument I mentioned in the previous comment

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -692,7 +727,10 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
                     << "    eps_p+eps_strain: " << (pressure_dilatation_based_dissipation_rate + strain_rate_tensor_based_dissipation_rate);
     }
     if(do_calculate_numerical_entropy){
-        this->pcout << "    Num. Entropy: " << std::setprecision(16) << numerical_entropy;
+        this->pcout << "    Num. Entropy cumulative, FR corrected: " << std::setprecision(16) << this->cumulative_numerical_entropy_change_FRcorrected;
```
jbrillon, line 730:
```suggestion
        this->pcout << "    Num. Entropy cumulative, FR corrected: " << std::setprecision(16) << this->get_numerical_entropy(); 
```

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -683,6 +717,7 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
         unsteady_data_table->write_text(unsteady_data_table_file);
     }
     // Print to console
+
```
jbrillon, line 720:
```suggestion
```

jbrillon: for avoiding merge conflicts

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -726,6 +764,7 @@ void PeriodicTurbulence<dim, nstate>::compute_unsteady_data_and_write_to_table(
             }
         }
     }
+
```
jbrillon, line 767:
```suggestion
```

jbrillon: to avoid merge conflicts

src/flow_solver/flow_solver_cases/periodic_turbulence.h
```diff
@@ -151,6 +169,7 @@ class PeriodicTurbulence : public PeriodicCubeFlow<dim,nstate>
 
     /// Data table storing the exact output times for the velocity field files
     std::shared_ptr<dealii::TableHandler> exact_output_times_of_velocity_field_files_table;
+
```
jbrillon, line 172:
```suggestion
```

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
jbrillon: or just remove the function since it doesn't look like its being called anywhere

src/flow_solver/flow_solver_cases/periodic_turbulence.h
```diff
@@ -102,7 +111,7 @@ class PeriodicTurbulence : public PeriodicCubeFlow<dim,nstate>
     bool is_taylor_green_vortex = false; ///< Identifies if taylor green vortex case; initialized as false.
     bool is_decaying_homogeneous_isotropic_turbulence = false; ///< Identified if DHIT case; initialized as false.
     bool is_viscous_flow = true; ///< Identifies if viscous flow; initialized as true.
-    bool do_calculate_numerical_entropy = false; ///< Identifies if numerical entropy should be calculated; initialized as false.
+    bool do_calculate_numerical_entropy=false; ///< Identifies if numerical entropy should be calculated; initialized as false.
```
jbrillon, line 114:
```suggestion
    bool do_calculate_numerical_entropy = false; ///< Identifies if numerical entropy should be calculated; initialized as false.
```

jbrillon: to avoid merge conflicts

src/flow_solver/flow_solver_cases/periodic_turbulence.h
cpethrick: That might be possible, but I would need time to think about how best to organize it. I may implement that the next time that I am making significant changes to this part of the code.

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
cpethrick: Per in-person discussion, this function is used in a test. I will resolve comments related to it.

src/flow_solver/flow_solver_cases/periodic_turbulence.h
cpethrick: Per our discussion in the lab, I think that reorganizing the `flow_solver_case` structure for entropy tests is a good idea. It won't get done on this PR, but is on my to-do list for the next time I need to work on this part of the code.

jbrillon: Sounds good, just reference this comment in a new issue and you can resolve this

cpethrick:
I re-ran ctest because I made changes to `flow_solver.cpp` and `ode_solver_base`. I've attached the log which confirms that tests are passing as expected. I ran on 4 processes and was using my computer for other tasks at the same time, therefore some timing tests have failed.

Also added a test for isentropic vortex convergence order using RRK because I realized that there was no test for `periodic_entropy_tests`.

src/flow_solver/flow_solver_cases/periodic_turbulence.h
cpethrick: https://github.com/dougshidong/PHiLiP/pull/195/files/c07024424c16a0aece8bda24603add99962cb3f4#r1523380835

jbrillon (approved):
LGTM, thanks for the PR!

PR #202, opened 2023-02-22
Replace `use_collocated_nodes` by `flux_nodes_type`

jbrillon:
PR to replace the `use_collocated_nodes` parameter by `flux_nodes_type`; allowing for over-integration with either GLL or GL flux nodes. For collocation, set `flux_nodes_type = GLL` and `overintegration=0` in the `.prm` file.

PR #203, opened 2023-02-24
Updating DHIT parameter files 

jbrillon:
PR to update the DHIT prm files that were mistakenly not updated right after PR #202 went through. Tests will fail without this update. Sorry for not catching this ahead of time.


PranshulThakur (approved):
LGTM! Thanks for catching it.

cpethrick (approved):
Thanks for making this change.

PR #207, opened 2023-03-09
Bugfix for writing higher-order data in 1D

cpethrick:
The changes in PR #181 introduced a bug where higher-order 1D results would either crash paraview or produce obviously incorrect results. The change in this PR yields nice, smooth plot on paraview.

For future reference, here is how to plot higher-order 1D results:
- Use the parameter `enable_higher_order_vtk_output = true` (it is false by default)
- There's no need to increase `grid_degree` to match `poly_degree`. 
- Open in paraview and use the filter PlotOverLine.
- Paraview will automatically use the higher-order result. No need to set `nonlinear subdivision level`.

Before, 16 elements with p=7:

After:


PranshulThakur (approved):
LGTM!

PR #201, opened 2023-02-17
Add isentropic vortex advection and Kelvin-Helmholtz instability test cases

cpethrick:
I'm adding these two test cases at the same time as both use similar setups, just different initial conditions.

A new `flow_solver_case` has been added because it doesn't make sense to have these Euler-based tests use `periodic_turbulence`. This structure is still under revision and I might make changes/rename/etc. 

The isentropic vortex text case was developed on PR#195, but I need the `periodic_entropy_tests` flow solver case for development on the Kelvin Helmholtz instability. Therefore, I've moved all changes to this branch such that isentropic vortex and KHI can be added together.

PR #209, opened 2023-03-14
KHI animation

cpethrick:
Extending from PR #201, this PR adds the KHI animation to the `README.md` file. The animation file size has been compressed enough that it is stored in a new folder, docs/gallery. We can use that folder for any other pictures we add.

It looks fine on desktop. When I'm on mobile, though, the KHI animation ends up being much bigger than the other two. I'm not sure how to fix that from a quick google. Happy to take any feedback regarding formatting - I'm not too familiar with markdown.


dougshidong (approved):
Lgtm

PR #210, opened 2023-03-22
Correction for NSFR with GL flux nodes for viscous TGV

jbrillon:
This PR is to incorporate a fix so that NSFR schemes with GL flux nodes is now stable when running the viscous TGV. The figure below shows the dissipation rate before ("cDG NSFR uncoll. [OLD]") and after ("cDG NSFR uncoll. [NEW]") the fix. Thank you to @AlexanderCicchino for finding and implementing this correction. In addition, I have added integration tests to go along with this case for both cases of flux nodes: GL and GLL.

image


cpethrick (approved):
The fix looks good to me and your results look much more stable.

Could you clarify the difference between using `soln_state_int/ext` and `soln_interp_to_face_int`? Was the old version interpolated incorrectly, or was it just not at the surface for GLL nodes?

AlexanderCicchino:
@cpethrick  soln_state_int/ext uses the volume entropy variables projected to the face, then transformed to the conservative variables--these are essential for the convective numerical flux since the Tadmor shuffle condition is discretely satisfied with respect to the entropy variables. The fix, using soln_interp_to_face_int for just the viscous numerical flux interpolates the conservative solution modal coefficients to the face. Although we don't yet have it proved, my hunch for using those for the viscous numerical flux rather than the entropy projected ones comes from the application of the diffusion tensor for an entropy dissipative result alike Chandrashekar "Kinetic energy preserving and entropy stable finite volume schemes for compressible Euler and Navier-Stokes equations" Sections 5 and 6.

cpethrick:
Thanks for the explanation @AlexanderCicchino!

PR #172, opened 2022-09-16
Reynolds averaged navier stokes sa neg

cyanmcg:
* Created Reynolds-averaged Navier-Stokes class derived from ModelBase
* Created Reynolds-averaged Navier-Stokes SA negative class derived from Reynolds-averaged Navier-Stokes
* Completed functions in physics_model for nstate > nstate_baseline 
* Added physical_source_term function in all PhysicsBase class

PR #212, opened 2023-03-31
Lax-Friedrichs flux correction

jbrillon:
Using `max_convective_eigenvalue` in `LaxFriedrichsRiemannSolverDissipation<dim,nstate,real>::evaluate_riemann_solver_dissipation()`. Implemented for all physics with `nstate=dim+2` (Euler, NS, RANS); others use the definition in `PhysicsBase` which returns `max_convective_eigenvalue` to keep the same behavior for those classes. 
- Reference:  Section 3.1 of Bernardo Cockburn, and Chi-Wang Shu,  "The Runge–Kutta Discontinuous Galerkin Method for Conservation Laws V",   JOURNAL OF COMPUTATIONAL PHYSICS 141, 199–224 (1998).

### DIFF
--- a/src/physics/euler.cpp
+++ b/src/physics/euler.cpp
@@ -928,7 +928,7 @@ real Euler<dim,nstate,real>
     const real sound = compute_sound (conservative_soln);
     real vel_dot_n = 0.0;
     for (int d=0;d<dim;++d) { vel_dot_n += vel[d]*normal[d]; };
-    const real max_normal_eig = sqrt(vel_dot_n*vel_dot_n) + sound;
+    const real max_normal_eig = abs(vel_dot_n) + sound;
 
     return max_normal_eig;
 }
--- a/src/physics/euler.cpp
+++ b/src/physics/euler.cpp
@@ -919,6 +919,22 @@ real Euler<dim,nstate,real>
 
 template <int dim, int nstate, typename real>
 real Euler<dim,nstate,real>
+::max_convective_normal_eigenvalue (
+    const std::array<real,nstate> &conservative_soln,
+    const dealii::Tensor<1,dim,real> &normal) const
+{
+    const dealii::Tensor<1,dim,real> vel = compute_velocities<real>(conservative_soln);
+
+    const real sound = compute_sound (conservative_soln);
+    real vel_dot_n = 0.0;
+    for (int d=0;d<dim;++d) { vel_dot_n += vel[d]*normal[d]; };
+    const real max_normal_eig = sqrt(vel_dot_n*vel_dot_n) + sound;
+
+    return max_normal_eig;
+}
+
+template <int dim, int nstate, typename real>
+real Euler<dim,nstate,real>
 ::max_viscous_eigenvalue (const std::array<real,nstate> &/*conservative_soln*/) const
 {
     return 0.0;
--- a/src/physics/reynolds_averaged_navier_stokes.cpp
+++ b/src/physics/reynolds_averaged_navier_stokes.cpp
@@ -325,6 +325,23 @@ real ReynoldsAveragedNavierStokesBase<dim,nstate,real>
 }
 //----------------------------------------------------------------
 template <int dim, int nstate, typename real>
+real ReynoldsAveragedNavierStokesBase<dim,nstate,real>
+::max_convective_normal_eigenvalue (
+    const std::array<real,nstate> &conservative_soln,
+    const dealii::Tensor<1,dim,real> &normal) const
+{
+    const std::array<real,nstate_navier_stokes> conservative_soln_rans = extract_rans_conservative_solution(conservative_soln);
+
+    const dealii::Tensor<1,dim,real> vel = this->navier_stokes_physics->template compute_velocities<real>(conservative_soln_rans);
+
+    real vel_dot_n = 0.0;
+    for (int d=0;d<dim;++d) { vel_dot_n += vel[d]*normal[d]; };
+    const real max_normal_eig = sqrt(vel_dot_n*vel_dot_n);
+
+    return max_normal_eig;
+}
+//----------------------------------------------------------------
+template <int dim, int nstate, typename real>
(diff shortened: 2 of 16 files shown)


jbrillon:
I will post the ctest results once I have them

PranshulThakur (approved):
LGTM! Thanks for catching the error.

src/physics/euler.cpp
```diff
@@ -917,6 +917,22 @@ ::max_convective_eigenvalue (const std::array<real,nstate> &conservative_soln) c
...
+::max_convective_normal_eigenvalue (
+    const std::array<real,nstate> &conservative_soln,
+    const dealii::Tensor<1,dim,real> &normal) const
+{
+    const dealii::Tensor<1,dim,real> vel = compute_velocities<real>(conservative_soln);
+
+    const real sound = compute_sound (conservative_soln);
+    real vel_dot_n = 0.0;
+    for (int d=0;d<dim;++d) { vel_dot_n += vel[d]*normal[d]; };
+    const real max_normal_eig = sqrt(vel_dot_n*vel_dot_n) + sound;
```
cpethrick, line 931: Why not use `abs(vel_dot_n)`?

src/physics/reynolds_averaged_navier_stokes.cpp
```diff
@@ -325,6 +325,23 @@ ::max_convective_eigenvalue (const std::array<real,nstate> &conservative_soln) c
...
+    const std::array<real,nstate> &conservative_soln,
+    const dealii::Tensor<1,dim,real> &normal) const
+{
+    const std::array<real,nstate_navier_stokes> conservative_soln_rans = extract_rans_conservative_solution(conservative_soln);
+
+    const dealii::Tensor<1,dim,real> vel = this->navier_stokes_physics->template compute_velocities<real>(conservative_soln_rans);
+
+    real vel_dot_n = 0.0;
+    for (int d=0;d<dim;++d) { vel_dot_n += vel[d]*normal[d]; };
+    const real max_normal_eig = sqrt(vel_dot_n*vel_dot_n);
```
cpethrick, line 339: Should this have +c? I'm not sure whether it's intended to be the same as Euler or not.

cpethrick (commented):
Thanks for making this change! I am interested to see what impact changing from `max_eig  = sqrt(vel2) + c` to `max_eig  = abs(vel*normal) + c`is. I think it would be good to wait until we have results comparing the old and new LxF fluxes to merge this into master.

src/physics/reynolds_averaged_navier_stokes.cpp
jbrillon: i asked him the same question back in PR #172 and apparently not, its consistent with his max conv eig function

src/physics/euler.cpp
jbrillon: fixed

dougshidong:
Not a bug. I believe I had discussed it with @cpethrick . The current implementation is seen in other papers/books.

Is there any specific reason why we want to change this now?

In the current implementation we're taking the **maximum** possible eigenvalue throughout the entire cell. If we're using LxF, we're not trying to minimize dissipation, we're voting for stability with as much dissipation as we can think of. The advantage of the current implementation is that we're adding the maximum "reasonable" amount. With the proposed changed, we're neither most stable nor least dissipative.

At the end of the day, it won't matter too much, it's going to be overdissipative anyways. If the goal is to have less dissipation, we can also just pick a more accurate numerical flux. 

Cool tangent if we want LxF, we can use graph-viscosity! Fun examples of this:
https://www.dealii.org/developer/doxygen/deal.II/step_69.html
https://arxiv.org/pdf/2004.08503.pdf
Effectively, the common LxF is just identity matrix times a factor, where we can choose a better diffusion tensor!!! d/dx(A_ij*(uR-uL)_j) awfully looks like a diffusion term doesn't it.


Anyways, I'm not against this PR. I just want to understand the expectations from this change

src/physics/reynolds_averaged_navier_stokes.cpp
dougshidong: Mmmh, kinda agree with @cpethrick on this one. Why would Euler have "+c", but not NS. What if we're doing low-Mach cases... then +c becomes the dominant term.

jbrillon: @cyanmcg

cyanmcg: For RANS, the max eigenvalues based on its flux Jacobian should be u since their convective fluxes always take the form of u_x*variable_turbulence (or u_y*variable_turbulence  or u_z*variable_turbulence), so the wave that carries this turbulent characteristic is the one with the flow velocity. And all other terms are considered as in the source, which does not contribute to RANS fluxes.

cyanmcg: @dougshidong the RANS class in the code only describes the additional PDE(s) for RANS turbulence models, which is derived from the model class, the RANS part (here I think the word using may lead confusion, I mean RANS in our code is just for the RANS turbulence model but not Reynolds-averaged Navier-Stokes + turbulence model) still using describe by the NS in the code, and as I last checked its function inherits from Euler should return the max convective eigenvalue as u+c. The RANS turbulence model only contributes to the dissipative flux, I think it may need some change to consider this part in the dissipative flux but not the convective flux

dougshidong: Makes a lot of sense.

dougshidong:
Up to you! The whole comment was more of a cautionary feedback in case you expected something "better" from this change.

dougshidong (approved):
Approving in case previous comment was construed as push back. Feel free to merge or not merge it.

cpethrick (approved):
Approving - as discussed with @jbrillon, it would be nice to see a direct comparison of the "maximum eigenvalue" and "normal eigenvalue" cases, but it isn't worth the computational resources as the new version yields results which are as expected, and is supported by literature.

PR #214, opened 2023-04-04
Better non-physical value handling and adding parmeters object to Physics

cpethrick:
A recent change prints warning messages whenever a negative density/pressure is detected in Euler. This is helpful for some tests, but results in very long log files for others (for instance, where an iterative solver sometimes fails but the test is built to handle this). Therefore, it is useful to add a parameter which will give the user more control over behaviour in these cases.

A function has been added to PhysicsBase such that it could be used to handle non-physical values in any PDE. It is only called in Euler, but other PDEs should be modified to use it in the future.

As an impact of this change, many constructors needed to be changed. Rather than adding a single new parameter, I have added a parameters pointer to PhysicsBase. I think this is a good structural addition. Constructors in `physics_factory` and `model_factory` have gotten very complex, and having a parameters object could allow us to simplify these constructors in the future.

Note that a dummy parameters pointer had to be added to all physics unit tests to be compatible with the changes to constructors.

CTEST passes as before.


cpethrick:
Marking this as ready for review.

After I have received and responded to feedback, I intend to open two issues for future work:
1. Simplify constructors in physics to take only the params object and read relevant data from params, rather than passing select values in the factory. Possibly overload the constructor for compatibility with old tests (especially the unit tests which were modified in this PR).
2. Add error-handling to other physics to catch NaN or non-physical negative values.

Both should be fairly straightforward changes, but are out of the scope of this PR.

src/parameters/all_parameters.cpp
```diff
@@ -267,6 +267,16 @@ void AllParameters::declare_parameters (dealii::ParameterHandler &prm)
...
+    prm.declare_entry("non_physical_behavior", "do_nothing",
+                      dealii::Patterns::Selection("do_nothing | abort_run | print_warning"),
+                      "Behavior when a nonphysical result is detected in physics, "
+                      "For example negative density or NaN. "
+                      "do_nothing will return BIG_NUMBER without any warnings "
+                      "abort_run will std::abort() "
+                      "print_warning will print a warning to console. "
+                      "Choices are <do_nothing | abort_run | print_warning>.");
+
+
```
jbrillon, line 279:
```suggestion
```

src/parameters/all_parameters.cpp
```diff
@@ -424,6 +433,12 @@ void AllParameters::parse_parameters (dealii::ParameterHandler &prm)
     if (flux_reconstruction_aux_string == "kPlus")       { flux_reconstruction_aux_type = kPlus; }
     if (flux_reconstruction_aux_string == "k10Thousand") { flux_reconstruction_aux_type = k10Thousand; }
 
+    const std::string non_physical_behavior_string = prm.get("non_physical_behavior");
+    if (non_physical_behavior_string == "do_nothing")       { non_physical_behavior_type = do_nothing;}
+    if (non_physical_behavior_string == "abort_run")        { non_physical_behavior_type = abort_run;}
+    if (non_physical_behavior_string == "print_warning")    { non_physical_behavior_type = print_warning;}
+
+
```
jbrillon, line 441:
```suggestion
```

src/physics/burgers_rewienski.cpp
```diff
@@ -7,14 +7,16 @@ namespace Physics {
 
 template <int dim, int nstate, typename real>
 BurgersRewienski<dim,nstate,real>::BurgersRewienski(
+        const Parameters::AllParameters *const                    parameters_input,
```
jbrillon, line 10:
```suggestion
        const Parameters::AllParameters *const parameters_input,
```

src/physics/physics.cpp
```diff
@@ -197,6 +202,24 @@ ::post_get_needed_update_flags () const
...
+::handle_non_physical_result() const
+{
+    if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::abort_run) {
+        this->pcout << "ERROR: Non-physical result has been detected. Aborting... " << std::endl << std::flush;
+        std::abort();
+    } else if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::print_warning) {
+        this->pcout << "WARNING: Non-physical result has been detected at a node." << std::endl;
+    } else if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::do_nothing) {
+        // do nothing -- assume that the test or iterative solver can handle this.
+    }
```
jbrillon, line 217: maybe add something like `const std::string non_physical_qty_name` as an argument to this function so that it prints what quantity is nonphysical like it did before in Euler

src/physics/physics.cpp
```diff
@@ -197,6 +202,24 @@ ::post_get_needed_update_flags () const
...
+template<typename real2>
+real2 PhysicsBase<dim,nstate,real>
+::handle_non_physical_result() const
+{
+    if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::abort_run) {
+        this->pcout << "ERROR: Non-physical result has been detected. Aborting... " << std::endl << std::flush;
+        std::abort();
+    } else if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::print_warning) {
+        this->pcout << "WARNING: Non-physical result has been detected at a node." << std::endl;
+    } else if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::do_nothing) {
```
jbrillon, line 215: maybe `set_to_big_number` instead of `do_nothing` ?

src/physics/euler.cpp
```diff
@@ -391,14 +378,8 @@ inline real2 Euler<dim,nstate,real>
...
-        entropy = log( entropy );
-    } else {
-        entropy = BIG_NUMBER;
-        this->pcout << "WARNING: Entropy is not defined because " << std::endl 
-                    << "    pressure * pow(density, -gam) < 0 ." << std::endl
-                    << "    Setting entropy = BIG_NUMBER." << std::endl;
-    }
+    check_positive_quantity<real2>(entropy);
+    entropy = log(entropy);
     return entropy;
```
jbrillon, line 383: did you intend to change the behavior? because now it will return `log(BIG_NUMBER)` instead of `BIG_NUMBER`

src/physics/physics.h
```diff
@@ -188,6 +211,7 @@ class PhysicsBase
      *  we should have a stable diffusive system
      */
     dealii::Tensor<2,dim,double> diffusion_tensor;
+
```
jbrillon, line 214:
```suggestion
```

src/physics/physics.h
```diff
@@ -176,6 +187,18 @@ class PhysicsBase
...
+    /** This is to be called by derived physics classes
+     *  when a nonphysical quantity is detected there.
+     *  e.g., negative density in Euler.
+     *  The behavior is determined by the value of
+     *  non_physical_behavior_type.
+     *  Returns BIG_NUMBER.
+     */
+    template<typename real2> 
+    real2 handle_non_physical_result () const;
+
```
jbrillon, line 201:
```suggestion
```

jbrillon (commented):
LGTM! Left a few comments, thanks for doing this tedious task!

src/physics/euler.cpp
jbrillon: which will be 100

cpethrick: Hmm, I didn't catch this. The `do_nothing` case will behave quite differently now...

cpethrick: One way I could modify the code is by adding (or modifying) `check_real_quantity()`. Then: 
* `entropy = log(entropy)`.  log(0) returns `-inf`, log(negative) returns `-nan`.
* `entropy = check_real_quantity(entropy)`, which would detect `nan` or `inf` and then pass to `this->handle_nonphysical_result()`.

src/physics/physics.cpp
cpethrick: Good suggestion. I've made a change which prints an error message.

src/physics/physics.cpp
cpethrick: Sure, that is more clear. Updated.

src/parameters/all_parameters.cpp
```diff
@@ -428,6 +437,11 @@ void AllParameters::parse_parameters (dealii::ParameterHandler &prm)
     if (flux_reconstruction_aux_string == "kPlus")       { flux_reconstruction_aux_type = kPlus; }
     if (flux_reconstruction_aux_string == "k10Thousand") { flux_reconstruction_aux_type = k10Thousand; }
 
+    const std::string non_physical_behavior_string = prm.get("non_physical_behavior");
+    if (non_physical_behavior_string == "return_big_number")       { non_physical_behavior_type = return_big_number;}
```
PranshulThakur, line 441: Consider replacing these assignments with 
`non_physical_behavior_type = NonPhysicalBehaviorEnum::return_big_number`... etc.

src/physics/euler.cpp
```diff
@@ -296,12 +287,7 @@ template <int dim, int nstate, typename real>
 inline real Euler<dim,nstate,real>
 ::compute_entropy_measure ( const real density, const real pressure ) const
 {
-    if (density < 0.0) {
```
PranshulThakur, line 299: Is it always the case that `compute_entropy_measure ( const real density, const real pressure ) ` is called through `compute_entropy_measure ( const std::array<real,nstate> &conservative_soln )` and not directly?

src/physics/physics.cpp
```diff
@@ -197,6 +202,31 @@ ::post_get_needed_update_flags () const
...
+{
+    if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::abort_run) {
+        this->pcout << "ERROR: Non-physical result has been detected. ";
+        if (message != "") {
+            this->pcout << std::endl << "    Message: " << message << std::endl;
+        }
+        this->pcout << " Aborting... " << std::endl << std::flush;
+        std::abort();
+    } else if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::print_warning) {
+        this->pcout << "WARNING: Non-physical result has been detected at a node." << std::endl;
```
PranshulThakur, line 218: Will it be better to output these statements through `std::cout`?

src/physics/euler.cpp
PranshulThakur: An if statement `if(entropy != BIG_NUMBER) { entropy = log(entropy);}` after calling `check_positive_quantity<real2>(entropy, "entropy");` could probably work.

PranshulThakur (commented):
Thanks for including various options to handle non-physical values. I have added a few comments.

src/physics/euler.cpp
PranshulThakur: An if statement `if(entropy != BIG_NUMBER) { entropy = log(entropy);}` after calling `check_positive_quantity<real2>(entropy, "entropy");` could probably work.

src/physics/euler.cpp
cpethrick: I'm not sure that we always use the `compute_entropy_measure ( const std::array<real,nstate> &conservative_soln )` form. It does make more sense to move the check to `compute_entropy_measure ( const real density, const real pressure )`.

This also falls under @jbrillon 's comment about changing the behaviour of the entropy check, which I'm still thinking about. I will fix this after addressing that comment.

src/parameters/all_parameters.cpp
cpethrick: For sure. I hadn't thought of doing that, but it is a little more clear. Fixed.

src/physics/euler.cpp
cpethrick: I have modified `check_positive_quantity` such that it returns a bool, which is used to return BIG_NUMBER if needed. Thanks @PranshulThakur for the suggestion!

I think this makes the behaviour the same as before. However it does lead to if-statements for nan detection in euler.cpp, which this PR sought to avoid. Let me know if there are any further suggestions for improvement.

src/parameters/all_parameters.cpp
```diff
@@ -267,6 +267,15 @@ void AllParameters::declare_parameters (dealii::ParameterHandler &prm)
                       "Dissipative numerical flux. "
                       "Choices are <symm_internal_penalty | bassi_rebay_2 | central_visc_flux>.");
 
+    prm.declare_entry("non_physical_behavior", "return_big_number",
+                      dealii::Patterns::Selection("return_big_number | abort_run | print_warning"),
+                      "Behavior when a nonphysical result is detected in physics, "
+                      "For example negative density or NaN. "
+                      "return_big_number will return BIG_NUMBER without any warnings "
```
jbrillon, line 274:
```suggestion
                      "return_big_number will return BIG_NUMBER (i.e. will set density to BIG_NUMBER) without any warnings "
```

cpethrick: I've changed the comment to "return_big_number will set the quantity to BIG_NUMBER without any warnings", since density is not the only value that is changed.

src/physics/physics.cpp
```diff
@@ -197,6 +202,31 @@ ::post_get_needed_update_flags () const
...
+    if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::abort_run) {
+        std::cout << "ERROR: Non-physical result has been detected. ";
+        if (message != "") {
+            std::cout << std::endl << "    Message: " << message << std::endl;
+        }
+        std::cout << " Aborting... " << std::endl << std::flush;
+        std::abort();
+    } else if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::print_warning) {
+        std::cout << "WARNING: Non-physical result has been detected at a node." << std::endl;
+        if (message != "") {
```
jbrillon, line 219:
```suggestion
        if (!message.empty()) {
```

jbrillon: could also move this after the else statement to reduce repetition

jbrillon: nevermind, not possible with the std::abort

src/physics/physics.cpp
```diff
@@ -197,6 +202,31 @@ ::post_get_needed_update_flags () const
...
 }
 
+template <int dim, int nstate, typename real>
+template<typename real2>
+real2 PhysicsBase<dim,nstate,real>
+::handle_non_physical_result(const std::string message) const
+{
+    if (this->non_physical_behavior_type == NonPhysicalBehaviorEnum::abort_run) {
+        std::cout << "ERROR: Non-physical result has been detected. ";
+        if (message != "") {
```
jbrillon, line 212:
```suggestion
        if (!message.empty()) {
```

src/physics/euler.cpp
```diff
@@ -152,20 +153,18 @@ ::convective_source_term (
...
 template<typename real2>
-void Euler<dim,nstate,real>::check_positive_density(real2 &density) const {
-    if (density < 0.0) {
-        this->pcout << "WARNING: Negative density encountered! Setting density as BIG_NUMBER." << std::endl;
-        density = BIG_NUMBER;
+bool Euler<dim,nstate,real>::check_positive_quantity(real2 &qty, const std::string qty_name) const {
+    bool qty_is_positive;
+    if (qty < 0.0) {
+        // Refer to base class for non-physical results handling
+        // Compiler complains unless we explicitly specify "template"
```
jbrillon, line 160:
```suggestion
```

jbrillon: this is standard when you have templated member functions in a parent class; you'll see it all throughout NavierStokes

src/physics/euler.cpp
jbrillon: looks good to me

src/physics/euler.cpp
```diff
@@ -296,12 +292,11 @@ template <int dim, int nstate, typename real>
...
-        this->pcout << "WARNING: Negative density encountered! Returning BIG_NUMBER as entropy measure." << std::endl;
-        return BIG_NUMBER;
-    } else {
-        return pressure*pow(density,-gam);
-    }
+    //Copy such that we don't modify the original density that is passed
+    real density_check = density;
+    const bool density_is_positive = check_positive_quantity<real>(density_check, "density");
+    if (density_is_positive)     return pressure*pow(density,-gam);
+    else                         return BIG_NUMBER;
```
jbrillon, line 299: Maybe move `BIG_NUMBER` to PhysicsBase as a public const variable since its declared twice now; once at the top of this file and once in the PhysicsBase function

src/physics/euler.cpp
```diff
@@ -390,16 +384,18 @@ template<typename real2>
...
-    real2 entropy = pressure * pow(density, -gam);
-    if (entropy > 0){
-        entropy = log( entropy );
+    // copy density and pressure such that the check will not modify originals
+    real2 density_check = density; 
+    real2 pressure_check = pressure;
+    const bool density_is_positive = check_positive_quantity(density_check, "density");
+    const bool pressure_is_positive = check_positive_quantity(pressure_check, "pressure");
+    if (density_is_positive && pressure_is_positive) {
+        real2 entropy = pressure * pow(density, -gam);
```
jbrillon, line 393: call `compute_entropy_measure` ?

src/physics/euler.cpp
```diff
@@ -390,16 +384,18 @@ template<typename real2>
...
+        entropy = log(entropy);
+        return entropy;
     } else {
-        entropy = BIG_NUMBER;
-        this->pcout << "WARNING: Entropy is not defined because " << std::endl 
-                    << "    pressure * pow(density, -gam) < 0 ." << std::endl
-                    << "    Setting entropy = BIG_NUMBER." << std::endl;
+        return BIG_NUMBER;
```
jbrillon, line 397: also, shouldn't the function in physics base be handling all instances where we set it to `BIG_NUMBER`?

cpethrick: In principle, I agree. However, this was the best way I could think of to replicate the original behaviour of the code. I could call `handle_non_physical_result` from this else clause, but then it will print warnings twice.

Let me know if you have ideas for how to execute this better.

src/physics/euler.cpp
cpethrick: I would rather leave that explicitly written. To me it's confusing to refer to "entropy measure" for this simple equation.

src/physics/euler.cpp
cpethrick: Sure, see recent commit.

cpethrick:
I ran ctest on 6 processes locally and most tests are passing. The results are attached. Please note that this was run one commit behind master, so the following testfails have been fixed by PR #224: 
	174 - 1D_BURGERS_STABILITY_ENERGY_LONG (Failed)
	177 - MPI_3D_EULER_SPLIT_TAYLOR_GREEN (Failed)
	178 - MPI_3D_EULER_SPLIT_TAYLOR_GREEN_CURV (Failed)
	182 - MPI_2D_ADVECTION_EXPLICIT_PERIODIC_ENERGY_WEIGHT_ADJUSTED_ON_THE_FLY_LONG (Failed)
I've ran each of these individually after merging to the current master and they pass as expected.

[ctestlog.txt](https://github.com/dougshidong/PHiLiP/files/11637671/ctestlog.txt)

PR #217, opened 2023-04-24
Flow Solver Memory Allocation Fix

jbrillon:
PR to incorporate fixes that @AlexanderCicchino discovered in terms of DG's and explicit ODE solver's allocation. This fix also now passes the necessary flags to `allocate_system` within Flow Solver. On the current version of master, for a viscous TGV simulation a total of 4^3 elements, the maximum polynomial degree I could get the code to allocate/run **locally** (32GB of ram, 8 processors), was **P5**. With these changes, I am now able to run **P15** with the same number of elements on the same computer (32GB of ram, 8 processors). This is expected to solve the memory allocation issues I've been having on the cluster.

### DIFF
--- a/src/dg/dg.cpp
+++ b/src/dg/dg.cpp
@@ -2095,7 +2095,7 @@ void DGBase<dim,real,MeshType>::allocate_system (
 
     dof_handler.distribute_dofs(fe_collection);
     //This Cuthill_McKee renumbering for dof_handlr uses a lot of memory in 3D, is there another way?
-    dealii::DoFRenumbering::Cuthill_McKee(dof_handler,true);
+    if(all_parameters->do_renumber_dofs) dealii::DoFRenumbering::Cuthill_McKee(dof_handler,true);
     //const bool reversed_numbering = true;
     //dealii::DoFRenumbering::Cuthill_McKee(dof_handler, reversed_numbering);
     //const bool reversed_numbering = false;
--- a/src/dg/dg.cpp
+++ b/src/dg/dg.cpp
@@ -2152,13 +2152,13 @@ void DGBase<dim,real,MeshType>::allocate_system (
     allocate_auxiliary_equation ();
 
     // System matrix allocation
-    dealii::DynamicSparsityPattern dsp(locally_relevant_dofs);
-    dealii::DoFTools::make_flux_sparsity_pattern(dof_handler, dsp);
-    dealii::SparsityTools::distribute_sparsity_pattern(dsp, dof_handler.locally_owned_dofs(), mpi_communicator, locally_relevant_dofs);
-
-    sparsity_pattern.copy_from(dsp);
-
     if (compute_dRdW || compute_dRdX || compute_d2R) {
+        dealii::DynamicSparsityPattern dsp(locally_relevant_dofs);
+        dealii::DoFTools::make_flux_sparsity_pattern(dof_handler, dsp);
+        dealii::SparsityTools::distribute_sparsity_pattern(dsp, dof_handler.locally_owned_dofs(), mpi_communicator, locally_relevant_dofs);
+
+        sparsity_pattern.copy_from(dsp);
+        
         system_matrix.reinit(locally_owned_dofs, sparsity_pattern, mpi_communicator);
     }
 
--- a/src/flow_solver/flow_solver.cpp
+++ b/src/flow_solver/flow_solver.cpp
@@ -42,8 +42,8 @@ FlowSolver<dim, nstate>::FlowSolver(
 , dg(DGFactory<dim,double>::create_discontinuous_galerkin(&all_param, poly_degree, flow_solver_param.max_poly_degree_for_adaptation, grid_degree, flow_solver_case->generate_grid()))
 {
     flow_solver_case->set_higher_order_grid(dg);
-    if (ode_param.allocate_AD_matrices) {
-        pcout << "Note: Allocating DG with AD matrices." << std::endl;
+    if (ode_param.allocate_matrix_dRdW) {
+        pcout << "Note: Allocating DG with AD matrix dRdW only." << std::endl;
         dg->allocate_system(true,false,false); // FlowSolver only requires dRdW to be allocated
     } else {
         pcout << "Note: Allocating DG without AD matrices." << std::endl;
--- a/src/flow_solver/flow_solver.cpp
+++ b/src/flow_solver/flow_solver.cpp
@@ -42,10 +42,12 @@ FlowSolver<dim, nstate>::FlowSolver(
 , dg(DGFactory<dim,double>::create_discontinuous_galerkin(&all_param, poly_degree, flow_solver_param.max_poly_degree_for_adaptation, grid_degree, flow_solver_case->generate_grid()))
 {
     flow_solver_case->set_higher_order_grid(dg);
-    if (flow_solver_param.steady_state == false) {
-        dg->allocate_system(false,false,false);
+    if (ode_param.allocate_AD_matrices) {
+        pcout << "Note: Allocating DG with AD matrices." << std::endl;
+        dg->allocate_system(true,false,false); // FlowSolver only requires dRdW to be allocated
     } else {
-        dg->allocate_system(true,false,false);
+        pcout << "Note: Allocating DG without AD matrices." << std::endl;
+        dg->allocate_system(false,false,false);
     }
 
     if(ode_param.ode_solver_type == Parameters::ODESolverParam::pod_galerkin_solver || ode_param.ode_solver_type == Parameters::ODESolverParam::pod_petrov_galerkin_solver){
--- a/src/flow_solver/flow_solver.cpp
+++ b/src/flow_solver/flow_solver.cpp
@@ -45,7 +45,7 @@ FlowSolver<dim, nstate>::FlowSolver(
     if (flow_solver_param.steady_state == false) {
         dg->allocate_system(false,false,false);
     } else {
-        dg->allocate_system(true,true,true);
+        dg->allocate_system(true,false,false);
     }
 
     if(ode_param.ode_solver_type == Parameters::ODESolverParam::pod_galerkin_solver || ode_param.ode_solver_type == Parameters::ODESolverParam::pod_petrov_galerkin_solver){
--- a/src/flow_solver/flow_solver.cpp
+++ b/src/flow_solver/flow_solver.cpp
@@ -42,7 +42,11 @@ FlowSolver<dim, nstate>::FlowSolver(
 , dg(DGFactory<dim,double>::create_discontinuous_galerkin(&all_param, poly_degree, flow_solver_param.max_poly_degree_for_adaptation, grid_degree, flow_solver_case->generate_grid()))
 {
     flow_solver_case->set_higher_order_grid(dg);
-    dg->allocate_system();
+    if (flow_solver_param.steady_state == false) {
+        dg->allocate_system(false,false,false);
+    } else {
+        dg->allocate_system(true,true,true);
+    }
 
     if(ode_param.ode_solver_type == Parameters::ODESolverParam::pod_galerkin_solver || ode_param.ode_solver_type == Parameters::ODESolverParam::pod_petrov_galerkin_solver){
         std::shared_ptr<ProperOrthogonalDecomposition::OfflinePOD<dim>> pod = std::make_shared<ProperOrthogonalDecomposition::OfflinePOD<dim>>(dg);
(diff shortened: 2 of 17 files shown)


src/flow_solver/flow_solver.cpp
```diff
@@ -42,7 +42,11 @@ FlowSolver<dim, nstate>::FlowSolver(
 , dg(DGFactory<dim,double>::create_discontinuous_galerkin(&all_param, poly_degree, flow_solver_param.max_poly_degree_for_adaptation, grid_degree, flow_solver_case->generate_grid()))
 {
     flow_solver_case->set_higher_order_grid(dg);
-    dg->allocate_system();
+    if (flow_solver_param.steady_state == false) {
```
dougshidong, line 45: Both implicit and explicit methods can be used for steady state and unsteady problems. For example Implicit RK that @cpethrick is implementing. (I know she uses matrix-free, but we don't need to build the matrix because it's matrix-free, not because it's steady)

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -110,14 +110,14 @@ template <int dim, typename real, int n_rk_stages, typename MeshType>
 void RungeKuttaODESolver<dim,real,n_rk_stages,MeshType>::allocate_ode_system ()
 {
     this->pcout << "Allocating ODE system and evaluating inverse mass matrix..." << std::endl;
-    this->solution_update.reinit(this->dg->right_hand_side);
+    this->solution_update.reinit(this->dg->locally_owned_dofs, this->dg->ghost_dofs, this->mpi_communicator);
```
dougshidong, line 113: Why are those changes related to large memory allocation?

tests/integration_tests_control_files/taylor_green_vortex_integration/viscous_TGV_energy_check_strong_adaptive_time_step.prm
```diff
@@ -13,6 +13,9 @@ set use_weak_form = false
 # Note: this was added to turn off check_same_coords() -- has no other function when dim!=1
 set use_periodic_bc = true
 
+# memory allocation
```
dougshidong, line 16: Maybe a more descriptive comment? Imagine explaining this flag to a new grad student.

src/flow_solver/flow_solver.cpp
jbrillon: Ah okay, so we should be checking the ODE solver type here instead?

src/parameters/all_parameters.cpp
```diff
@@ -284,6 +284,10 @@ void AllParameters::declare_parameters (dealii::ParameterHandler &prm)
                       dealii::Patterns::Bool(),
                       "Outputs the surface solution vtk files. False by default");
 
+    prm.declare_entry("do_renumber_dofs", "true",
+                      dealii::Patterns::Bool(),
+                      "Flag for renumbering DOFs. True by default. Set to false if doing 3D unsteady flow simulations.");
```
cpethrick, line 289:
```suggestion
                      "Flag for renumbering DOFs using Cuthill-McKee renumbering. True by default. Set to false if doing 3D unsteady flow simulations.");
```

src/dg/dg.cpp
```diff
@@ -2095,7 +2095,7 @@ void DGBase<dim,real,MeshType>::allocate_system (
 
     dof_handler.distribute_dofs(fe_collection);
     //This Cuthill_McKee renumbering for dof_handlr uses a lot of memory in 3D, is there another way?
-    dealii::DoFRenumbering::Cuthill_McKee(dof_handler,true);
+    if(all_parameters->do_renumber_dofs) dealii::DoFRenumbering::Cuthill_McKee(dof_handler,true);
```
cpethrick, line 2098: Might be a question for @dougshidong - what is the purpose of this renumbering and why is it such a memory bottleneck?

src/flow_solver/flow_solver.cpp
cpethrick: We could add a parameter `allocate_AD_matrices` which is set `true` according to the user's chosen `ode_solver_type` (rather than adding a parameter which the user would specify explicitly). Or just add some logic here if such a parameter isn't useful elsewhere.

src/parameters/all_parameters.cpp
cpethrick: Maybe add a description to the effect of "... renumbering increases memory use..." so that a new student could understand easier when to set this.

Also, is this something which could be set according to already-existing parameters? I don't completely understand in what case we want renumbering. But if it's associated with AD runs, maybe we could use `ode_solver_type` as an indicator? I'm suggesting this because the more parameters we have, the more likely we are to make mistakes while setting them. (I know I've forgotten to change parameters many times, causing a useless result!)

cpethrick (commented):
Thanks for adding these memory fixes! I've added a few questions and suggestions.

src/flow_solver/flow_solver.cpp
jbrillon: Maybe each ODE class should have a bool allocate_AD_matrices that gets assigned in the ODE solver base constructor; similar to how in Physics we have the bool has_nonzero_dissipative_flux. That way there's no dependency on FlowSolver

src/ode_solver/explicit_ode_solver.cpp
AlexanderCicchino: https://www.dealii.org/current/doxygen/deal.II/group__Vectors.html#ga3be6c4ce529bb9b6c13eb831d0a86f55 
When initialized with the global size rather than the partitioning info, then it doesn't do any memory partitioning.

src/ode_solver/explicit_ode_solver.cpp
```diff
@@ -110,14 +110,14 @@ template <int dim, typename real, int n_rk_stages, typename MeshType>
 void RungeKuttaODESolver<dim,real,n_rk_stages,MeshType>::allocate_ode_system ()
 {
     this->pcout << "Allocating ODE system and evaluating inverse mass matrix..." << std::endl;
```
cpethrick, line 112:
```suggestion
    this->pcout << "Allocating ODE system..." << std::endl;
```

src/flow_solver/flow_solver.cpp
```diff
@@ -42,7 +42,11 @@ FlowSolver<dim, nstate>::FlowSolver(
...
+        dg->allocate_system(false,false,false);
+    } else {
+        dg->allocate_system(true,true,true);
```
PranshulThakur, line 48: We currently cannot compute dRdW, dRdX and second derivatives all at once. WeakDG uses codiPack and the corresponding functions (see, for example, [DGWeak::assemble_volume_term_derivatives()](https://github.com/dougshidong/PHiLiP/blob/f65960b8139e25f8e819a086261dbd605374d9f5/src/dg/weak_dg.cpp#L4130)) compute either first or second derivatives. For the flow solver, we just need residual derivative w.r.t. solution and this line can be replaced with `dg->allocate_system(true,false,false)`.

PranshulThakur (commented):
Thanks for improving memory allocations in the code.

src/ode_solver/explicit_ode_solver.cpp
dougshidong: `this->dg->right_hand_side` is not a global size, it is a vector.

It uses the `reinit` function akin to the copy constructor
https://www.dealii.org/current/doxygen/deal.II/group__Vectors.html#ga8a79ca37d7f7a8cfd04f5dd61d350ae5
such that the `solution_update` will have the same parallel partitioning as `right_hand_size`

Can we revert this change?

src/dg/dg.cpp
dougshidong: Big paragraph ahead. Simply because I'm procrastinating on work I should actually be doing.

### What it doesn't do

Reordering does *not* change inherently changes the conditioning of your matrix. Your eigenvalue should stay the same, but your eigenvectors will be switched around. It's the same problem, from a different angle. [1, 3; 2 9] has the same eigenvalues as [9 2; 3 1]

### What it does
What reodering does offer is a lower fill-in when performing ILU preconditioning.

### LU Factorization
#### Sidebar 1
ILU aims to approximate P~A^{-1} such that solving P*A x = P*b is "easy". ILU approximates A^{-1} by doing the usual LU factorization. However, it will only store values v_{i,j} in L and U if the value of A^k is non-zero (k being ILU(k+1)). The new values are called fill-in. It essentially tries as keep a semblance of the sparsity pattern of A. 
#### Sidebar 2
The bandwidth of a matrix is defined by the maximum distance an entry is from the diagonal. Aka, how tightly diagonal it is. The LU algorithm is in nature very sequential (unless you ask Aditya Kashi), such that the the operation you do on one row depends on the previous operation. Now, if you can imagine how keeping a low bandwidth propagates as much information as possible while not creating too many non-zeros. Great [video](https://www.youtube.com/watch?v=ymIvps7pgRk) that gives a feel for "information propagation" based on numbering.
#### Impact
Now, if A has a "random" sparsity pattern, the LU factorization of that matrix will not necessarily correspond to the sparsity pattern of the original matrix. Therefore, the incomplete factorization will likely drop many entries that would have otherwise been important. However, when reordering DoFs using Cuthill_McKee, we somewhat reduce the *bandwidth* of the matrix. Notice the many -ly and uncertain words above; this is because a lot of this is based on empirical tests and heuristics.


### Actually trying to answer your question instead of babbling

> this reorders the cells such that we have a block diagonal matrix

Cuthill_Mckee does not form block diagonals. It tries to minimize the bandwidth of the matrix. This is a heuristic to minimize graph bandwidth, it does not actually minimize it.

> which is essential for the conditioning of the linear solver

Yeah, kind of. It is "essential" specifically for ILU in CFD. However, it can also be used for LUSGS algorithms like in the video above that requires no matrix to be stored. Cuthill-McKee specifically is just a heuristic.

> Looping over all of the dofs is understandably memory-intensive.

Kind of false. Looping over the dofs take no memory. It's what you do inside of it. My understanding is that some variation of breadth-first search is done on the adjacency matrix. The adjacency matrix (likely implemented as adjacency list that will store as many entries as non-zeroes) will then require memory proportional to the actual sparse matrix.

### Reordering

Give this [link](https://www.dealii.org/current/doxygen/deal.II/namespaceDoFRenumbering.html) a read.

Reordering can be important for so much more than ILU. Sometimes, you may want to separate your equations, like density, momentum, and energy because you can use a specific preconditioner on momentum, and another one on density and energy. In this scenario, a component-wise reordering would be used. Or maybe you rely on the block-property of DG cells, because you want to apply a transformation with some underlying physics. Then you would use cell-wise numbering. You can even use some fancy reordering scheme in a way that numbers "emanate" from airfoil to farfield if you believe that physical information propagates that way. Or maybe you just want things to be close to each other on your computer such that accessing them is faster due to less cache misses.

### Takeaway

Renumbering has 0 impact on the result of explicit time-stepping because there is no dependency on ordering. However, it may still impact its memory-access efficiency.

Reordering is very important for any type of implicit schemes because it affects the order in which information is propagated from cell to cell.

dougshidong: Also, we should probably keep reordering on for any implicit schemes. There is little to no chance most iterative implicit schemes converge for a random ordering.

cpethrick: Thank you for the detailed explanation! I agree that my explanation was vague (and, as you pointed out, a bit incorrect in places!). I will read through the links you provided, it's always best to understand what's going on in the code.

src/flow_solver/flow_solver.cpp
jbrillon: The default is `(true,true,true)` (see `dg.h`) so that's what was being done before, should I still replace it by `(true,false,false)`?

src/ode_solver/explicit_ode_solver.cpp
AlexanderCicchino: Ok yes let's revert it. I'm no longer able to replicate the memory jump. The memory fix must've been from the commit before and I didn't see it from the build. Thanks!

src/flow_solver/flow_solver.cpp
PranshulThakur: Technically flow solver just needs dRdW and it's best to only allocate it's memory. It might help prevent memory issues when we run implicit solver on large test cases.

jbrillon: fixed

src/ode_solver/explicit_ode_solver.cpp
jbrillon: fixed

tests/integration_tests_control_files/taylor_green_vortex_integration/viscous_TGV_energy_check_strong_adaptive_time_step.prm
jbrillon: fixed

src/flow_solver/flow_solver.cpp
jbrillon: Nvm, this won't work since we create the ODE solver after having allocated DG

jbrillon: I'll go with a parameter approach, that gets assigned in the parameter reader based on the ODE solver type

jbrillon: fixed

sivanadarajah (approved):
Alex, Excellent work. I am glad that this was sorted out.

PR #219, opened 2023-05-11
Efficient implementation with Scaling for DG Strong

AlexanderCicchino:
This PR addresses:
1) Remaining memory issues for high polynomial orders for linear and curvilinear meshes.
2) Provides scaling for DG strong conservative, overintegrated and entropy conserving on a straight and curvilinear mesh. Showing the entire strong DG scales at O(n^{d+1}), up to 136^3 dofs ran locally on 4GB of ram.
3) Restructuring the two-point flux and creating a vector sparsity structure as to only compute O(n^{d+1}) terms on the fly and apply them.
4) Overintegrated entropy conserving schemes now conserve entropy at machine precision on general integration strength by introducing a new solution basis entropy projected variable.
Below, please find results for conservative strong DG, FR c+ with and without overintegration. Also, DG and c+ entropy conserving "EC" two-point flux form using Ranocha's two-point flux for both Cartesian and nonsymmetric curvilinear meshes.
[3D_TGV_scaling_curv_legend_right.pdf](https://github.com/dougshidong/PHiLiP/files/11490524/3D_TGV_scaling_curv_legend_right.pdf)
[3D_scaling_cartesian_TGV_legend_right.pdf](https://github.com/dougshidong/PHiLiP/files/11490528/3D_scaling_cartesian_TGV_legend_right.pdf)


src/dg/dg.cpp
```diff
@@ -2729,7 +2776,7 @@ void DGBase<dim,real,MeshType>::apply_inverse_global_mass_matrix(
     const FR_Aux_enum FR_Type_Aux = this->all_parameters->flux_reconstruction_aux_type;
      
     const unsigned int init_grid_degree = high_order_grid->fe_system.tensor_degree();
-    OPERATOR::mapping_shape_functions<dim,2*dim> mapping_basis(1, max_degree, init_grid_degree);
+    OPERATOR::mapping_shape_functions<dim,2*dim> mapping_basis(1, init_grid_degree, init_grid_degree);
```
jbrillon, line 2779: I'm not too familiar with this function, so I was wondering why it's changed from `max_degree` to `init_grid_degree`

AlexanderCicchino: Max degree is for the solution polynomial basis, whereas init grid degree is used for the metric mapping shape function. The mapping shape functions are used to formulate the metric Jacobian etc and all transformations between physical and reference spaces.

src/parameters/all_parameters.cpp
```diff
@@ -90,6 +90,14 @@ void AllParameters::declare_parameters (dealii::ParameterHandler &prm)
                       dealii::Patterns::Bool(),
                       "Use other boundary conditions by default. Otherwise use periodic (for 1d burgers only");
 
+    prm.declare_entry("renumber_dof_handler_Cuthill_Mckee", "true",
+                      dealii::Patterns::Bool(),
+                      "Renumber dof handler with Cuthill-Mckee by default. Don't renumber dof_handler if false.");
+
+    prm.declare_entry("use_curvilinear_grid", "false",
+                      dealii::Patterns::Bool(),
+                      "Use straight grid by default. Curvilinear is true.");
```
cpethrick, line 99:
```suggestion
                      "Use straight grid by default. Curvilinear is true. Only used in taylor_green_scaling test.");
```
I think this should eventually be moved out of `all_parameters` as it is a test-specific parameter. If we ever move the curvilinear TGV tests into flow solver, we should set it there. For now, clarifying where it is used is good enough for me.

src/testing/TGV_scaling.cpp
```diff
@@ -0,0 +1,381 @@
...
+            dealii::Triangulation<dim>::smoothing_on_refinement |
+            dealii::Triangulation<dim>::smoothing_on_coarsening));
+
+    using real = double;
+
+    PHiLiP::Parameters::AllParameters all_parameters_new = *all_parameters;  
+    double left = 0.0;
+    double right = 2 * dealii::numbers::PI;
+
+    const unsigned int n_refinements = all_parameters->flow_solver_param.number_of_grid_elements_per_dimension;
```
cpethrick, line 274: maybe flow_solver_param.number_of_mesh_refinements would be better?

src/testing/TGV_scaling.cpp
```diff
@@ -0,0 +1,381 @@
...
+        dealii::GridTools::collect_periodic_faces(*grid,2,3,1,matched_pairs);
+        dealii::GridTools::collect_periodic_faces(*grid,4,5,2,matched_pairs);
+        grid->add_periodicity(matched_pairs);
+        grid->refine_global(n_refinements);
+    }
+
+    std::ofstream myfile (all_parameters_new.energy_file + ".gpl"  , std::ios::trunc);
+    const unsigned int poly_degree_start= all_parameters->flow_solver_param.poly_degree;
+
+    const unsigned int poly_degree_end = 25;
```
cpethrick, line 294: Is it feasible to run this high on a local computer?

cpethrick (commented):
This new test will be great for understanding the memory performance of the unsteady part of the code. I am looking forward to using this test to understand memory in the ode solver! Great work, this will really improve our code as we run bigger cases.

I've left a few comments, mostly minor fixes and clarifying questions, and some comments on the new test. I haven't gone through dg or high_order grid yet.

src/operators/operators.cpp
```diff
@@ -809,6 +810,258 @@ void SumFactorizedOperators<dim,n_faces>::Hadamard_product(
     }
 }
 
+template <int dim, int n_faces>  
+void SumFactorizedOperators<dim,n_faces>::sum_factorized_Hadamard_sparsity_pattern(
```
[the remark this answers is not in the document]
AlexanderCicchino: Based on unpublished work, based off the basis structure in reference space, we can solve a HAdamard product in O(n^{d+1}) flops, similar to sum-factorization. The issue with the previous implementation was that it would allocate a matrix of size n^d x n^d then perform O(n^{d+1}) flops with it. When allocating a matrix of size n^d x n^d, that creates an opertaion of size O(n^{2d}). So the new functions creat a pseudo sparsity structure, where we know what the index positions would be in an n^d x n^d matrix without ever having to allocate n^2d. Instead, it allocates an n^d x n sized matrix and fills it with the non-zero elements. These relate to performance because it allows the solver to scale at O(n^{d+1}) completely, for both flops and allocation on-the-fly.

src/testing/TGV_scaling.cpp
```diff
@@ -0,0 +1,381 @@
...
+    std::shared_ptr<Triangulation> grid = std::make_shared<Triangulation>(
+        MPI_COMM_WORLD,
+        typename dealii::Triangulation<dim>::MeshSmoothing(
+            dealii::Triangulation<dim>::smoothing_on_refinement |
+            dealii::Triangulation<dim>::smoothing_on_coarsening));
+
+    using real = double;
+
+    PHiLiP::Parameters::AllParameters all_parameters_new = *all_parameters;  
+    double left = 0.0;
```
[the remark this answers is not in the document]
AlexanderCicchino: Could in the future but right now the test is only for TGV.

src/testing/TGV_scaling.cpp
AlexanderCicchino: Yes, on a local computer with 4GB of RAM to be used, I could go up to p=36 with 4^3 liner grid. The test goes up to p=25 because for the curvilinear case, there is still something in high_order_grid causing the memory to spike past p=25.

src/testing/TGV_scaling.cpp
```diff
@@ -0,0 +1,381 @@
...
+                        / std::log((double)((i+1.0)/(i)))<<
+        std::endl;
+        if(i>poly_degree_end-4){
+            avg_slope1 += std::log(((double)time_to_run[i]) /( (double)time_to_run_mpi[i-1]))
+                        / std::log((double)((i+1.0)/(i)));
+        }
+    }
+    avg_slope1 /= (3.0);
+
+    if(abs(dim + 1 - avg_slope1) > 0.1){
```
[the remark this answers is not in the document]
AlexanderCicchino: I am not sure. Before in different tests I used a different way to measure CPU time.

src/testing/euler_split_inviscid_taylor_green_vortex.cpp
```diff
@@ -243,16 +243,13 @@ double EulerTaylorGreen<dim, nstate>::get_timestep(std::shared_ptr < DGBase<dim,
         }
         const double max_eig = *(std::max_element(convective_eigenvalues.begin(), convective_eigenvalues.end()));
 
-        double cfl = 0.1 * delta_x/max_eig;
-
+        const double max_eig_mpi = dealii::Utilities::MPI::max(max_eig, mpi_communicator);
```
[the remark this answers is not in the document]
AlexanderCicchino: Yes it would, fortunately this test doesn't use the function. I will delete the unused functions.:)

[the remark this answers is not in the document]
AlexanderCicchino: Yes I will update this function from a different branch thank you.

src/mesh/high_order_grid.cpp
```diff
@@ -98,25 +102,33 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::initialize_with
...
+        // pcout << "Disabled check_valid_cells. Took too much time due to shape_grad()." << std::endl;
+        for (; cell!=endcell; ++cell) {
+            if (!cell->is_locally_owned())  continue;
+            const bool is_invalid_cell = check_valid_cell(cell);
+         
+            if ( !is_invalid_cell ) {
+                std::cout << " Poly: " << max_degree
+                          << " Grid: " << nth_refinement
+                          << " Cell: " << cell->active_cell_index() << " has an invalid Jacobian." << std::endl;
+                // bool fixed_invalid_cell = fix_invalid_cell(cell);
```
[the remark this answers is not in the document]
AlexanderCicchino: No, only DG strong will skip this part because it verifies the metric Jacboian determinant on the fly.

src/operators/operators.cpp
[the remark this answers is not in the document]
cpethrick: Perfect, thanks for the explanation!

src/testing/TGV_scaling.cpp
cpethrick: Wow! That's quite an improvement :) p=25 looked quite high from the reference of the previous capabilities.

src/dg/dg.cpp
jbrillon: Ah I see, thanks for the explanation

src/testing/TGV_scaling.cpp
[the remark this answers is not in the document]
cpethrick: Okay. I don't see why we would ever change the left and right bounds while using this test, so it's fine with me to leave it as-is!

src/testing/TGV_scaling.cpp
```diff
@@ -0,0 +1,381 @@
...
+        pcout<<"Poly Degree "<<poly_degree<<" time to run Mpi "<<std::fixed << std::setprecision(16) << (double)time_to_run_mpi[poly_degree]<<std::endl;
+        myfile << poly_degree << " " << std::fixed << std::setprecision(16) << time_to_run_mpi[poly_degree]<< std::endl;
+    }//end of poly loop
+
+    myfile.close();
+    double avg_slope1 = 0.0;
+    pcout<<"Times for one timestep"<<std::endl;
+    pcout<<"local time  | Slope |  "<<"MPI sum time | Slope "<<std::endl;
+    for(unsigned int i=poly_degree_start+1; i<poly_degree_end; i++){
+        pcout<<(double)time_to_run[i]<<" "<<std::log(((double)time_to_run[i]) / ((double)time_to_run[i-1]))
```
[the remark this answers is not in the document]
AlexanderCicchino: Correction: the CPUs actually take a different amount of time.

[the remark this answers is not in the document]
AlexanderCicchino: fixed.

src/testing/TGV_scaling.cpp
```diff
@@ -0,0 +1,381 @@
...
+    double right = 2 * dealii::numbers::PI;
+
+    const unsigned int n_refinements = all_parameters->flow_solver_param.number_of_grid_elements_per_dimension;
+    if(all_parameters->use_curvilinear_grid){
+        //if curvilinear
+        PHiLiP::Grids::nonsymmetric_curved_grid<dim,Triangulation>(*grid, n_refinements);
+    }
+    else{
+        //if straight
+        const bool colorize = true;
```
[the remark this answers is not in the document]
AlexanderCicchino: Fixed

tests/integration_tests_control_files/TGV_scaling/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
+##Linear grid cDG
```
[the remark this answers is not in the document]
AlexanderCicchino: They now take about 3 minutes long each. They are all important since they each consider a different aspect of the solver.

src/parameters/all_parameters.cpp
AlexanderCicchino: Ideally, there would be a mesh specific parameters where you could choose which type of grid and all of the grid parameters. For now will use your suggested commit.

src/parameters/all_parameters.cpp
```diff
@@ -127,6 +135,10 @@ void AllParameters::declare_parameters (dealii::ParameterHandler &prm)
                       dealii::Patterns::Bool(),
                       "Build global mass inverse matrix and apply it. Otherwise, use inverse mass on-the-fly by default for explicit timestepping.");
 
+    prm.declare_entry("check_valid_metric_Jacobian", "true",
+                      dealii::Patterns::Bool(),
+                      "Check validty of metric Jacobian when high-order grid is constructed by default. Do not check if false. Not checking is useful if the metric terms are built on the fly with operators, it reduces the memory cost for high polynomial grids. It is always false for strong form DG.");
```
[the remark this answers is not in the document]
AlexanderCicchino: Fixed as suggested.

src/testing/euler_split_inviscid_taylor_green_vortex.cpp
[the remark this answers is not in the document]
AlexanderCicchino: fixed.

src/testing/euler_split_inviscid_taylor_green_vortex.cpp
```diff
@@ -272,8 +269,10 @@ int EulerTaylorGreen<dim, nstate>::run_test() const
     PHiLiP::Parameters::AllParameters all_parameters_new = *all_parameters;  
     double left = 0.0;
     double right = 2 * dealii::numbers::PI;
-    const int n_refinements = 2;
-    unsigned int poly_degree = 3;
+//    const int n_refinements = 2;
+    const unsigned int n_refinements = all_parameters->flow_solver_param.number_of_grid_elements_per_dimension;
+//    unsigned int poly_degree = 3;
```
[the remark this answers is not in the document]
AlexanderCicchino: fixed

src/testing/euler_split_inviscid_taylor_green_vortex.cpp
```diff
@@ -307,67 +306,45 @@ int EulerTaylorGreen<dim, nstate>::run_test() const
...
-          return 1;
+    ode_solver->current_iteration = 0;
+    ode_solver->allocate_ode_system();
+
+    const double initial_energy = compute_kinetic_energy(dg, poly_degree);
+    const double initial_energy_mpi = (dealii::Utilities::MPI::sum(initial_energy, mpi_communicator));
+    while(ode_solver->current_time < finalTime){
+        const double time_step =  get_timestep(dg,poly_degree, delta_x);
+        if(ode_solver->current_iteration%all_parameters_new.ode_solver_param.print_iteration_modulo==0)
+            pcout<<"time step "<<time_step<<" current time "<<ode_solver->current_time<<std::endl;
```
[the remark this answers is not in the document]
AlexanderCicchino: fixed

src/testing/euler_split_inviscid_taylor_green_vortex.cpp
```diff
@@ -272,8 +269,10 @@ int EulerTaylorGreen<dim, nstate>::run_test() const
     PHiLiP::Parameters::AllParameters all_parameters_new = *all_parameters;  
     double left = 0.0;
     double right = 2 * dealii::numbers::PI;
-    const int n_refinements = 2;
-    unsigned int poly_degree = 3;
+//    const int n_refinements = 2;
```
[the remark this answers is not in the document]
AlexanderCicchino: fixed

src/testing/TGV_scaling.cpp
AlexanderCicchino: fixed

src/operators/operators.h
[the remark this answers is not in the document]
AlexanderCicchino: Maybe for a future PR? Could be raised as an issue if that's ok?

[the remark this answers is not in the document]
cpethrick: Sounds good!

src/testing/TGV_scaling.cpp
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        //store local cpu time
+        time_to_run[poly_degree] = dg->assemble_residual_time;
+        //store mpi summed cpu time
+        time_to_run_mpi[poly_degree] = dealii::Utilities::MPI::sum(time_to_run[poly_degree], this->mpi_communicator);
+        
+        std::cout<<"Poly Degree "<<poly_degree<<" time to run local cpu "<<std::fixed << std::setprecision(16) << (double)time_to_run[poly_degree]<<std::endl;
+        pcout<<"Poly Degree "<<poly_degree<<" time to run Mpi "<<std::fixed << std::setprecision(16) << (double)time_to_run_mpi[poly_degree]<<std::endl;
```
cpethrick, line 101: Maybe add an MPI barrier here - otherwise the MPI output line can print in the middle of the "local cpu" times.

src/dg/dg.cpp
```diff
@@ -2764,6 +2809,9 @@ void DGBase<dim,real,MeshType>::apply_inverse_global_mass_matrix(
         }
     }
 
+    dealii::Timer timer;
```
jbrillon, line 2812:
```suggestion
    dealii::Timer timer(this->mpi_communicator,false);
```

src/dg/dg.cpp
```diff
@@ -2143,13 +2182,19 @@ void DGBase<dim,real,MeshType>::allocate_system (
     //right_hand_side.reinit(locally_owned_dofs, mpi_communicator);
     right_hand_side.reinit(locally_owned_dofs, ghost_dofs, mpi_communicator);
     right_hand_side.add(1.0); // Avoid 0 initial residual for output and logarithmic visualization.
-    dual.reinit(locally_owned_dofs, ghost_dofs, mpi_communicator);
+
+    allocate_dual_vector();
+//    dual.reinit(locally_owned_dofs, ghost_dofs, mpi_communicator);
```
jbrillon, line 2187:
```suggestion
```

src/dg/dg.cpp
```diff
@@ -1474,16 +1502,25 @@ void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, cons
...
                 mapping_basis,
                 false,
                 right_hand_side,
                 auxiliary_right_hand_side);
         } // end of cell loop
+
+       // assemble_residual_time += clock() - time_start_cell_loop;
+       // assemble_residual_time += dealii::Utilities::MPI::max(timer.wall_time(), this->mpi_communicator);
+        timer.stop();
+       // assemble_residual_time += dealii::Utilities::MPI::sum(timer.cpu_time(), this->mpi_communicator);
```
jbrillon, line 1516: can remove the commented lines

src/dg/dg.cpp
```diff
@@ -3038,9 +3089,10 @@ void DGBase<dim,real,MeshType>::apply_global_mass_matrix(
                     projection_oper.matrix_vector_mult_1D(JxW,
                                                           temp,
                                                           projection_oper.oneD_vol_operator);
-                    mass.inner_product_1D(temp, ones,
-                                          local_output_vector,
-                                          mass.oneD_vol_operator);
+                   // mass.inner_product_1D(temp, ones,
```
jbrillon, line 3092:
```suggestion
```

src/dg/strong_dg.cpp
```diff
@@ -979,33 +1064,59 @@ void DGStrong<dim,nstate,real,MeshType>::assemble_volume_term_strong(
...
+            soln_state = this->pde_physics_double->compute_conservative_variables_from_entropy_variables (entropy_var);
+            
+            //loop over all the non-zero entries for "sum-factorized" Hadamard product that corresponds to the iquad.
+            for(unsigned int row_index = iquad * n_quad_pts_1D, column_index = 0; 
+               // Hadamard_rows_sparsity[row_index][0] == iquad; 
+                column_index < n_quad_pts_1D; 
+                row_index++, column_index++){
+
+                if(Hadamard_rows_sparsity[row_index][0] != iquad){
+                    pcout<<"The volume Hadamard rows sparsity pattern does not match."<<std::endl;
```
jbrillon, line 1082:
```suggestion
                    pcout<<"The volume Hadamard rows sparsity pattern does not match. Aborting..."<<std::endl;
```

src/dg/strong_dg.cpp
```diff
@@ -1295,7 +1434,9 @@ void DGStrong<dim,nstate,real,MeshType>::assemble_boundary_term_strong(
 
         // Evaluate physical convective flux
         std::array<dealii::Tensor<1,dim,real>,nstate> conv_phys_flux;
-        conv_phys_flux = this->pde_physics_double->convective_flux (soln_state);
+        if(!this->all_parameters->use_split_form && !this->all_parameters->use_curvilinear_split_form){
+            conv_phys_flux = this->pde_physics_double->convective_flux (soln_state);
+        }
```
jbrillon, line 1439: can you explain why this if statement is added now throughout this section? `conv_phys_flux` isn't needed when `use_split_form == true` and `use_curvilinear_split_form == false`?

src/dg/strong_dg.cpp
```diff
@@ -1339,11 +1484,12 @@ void DGStrong<dim,nstate,real,MeshType>::assemble_boundary_term_strong(
...
-    for(int idim=0; idim<dim; idim++){
-        if(std::abs(unit_ref_normal_int[idim]) >= 1e-11)//reference unit normal is 0 in all direction and 1 in outward normal direction
-            dim_not_zero = idim;//this is outward pointing normal reference direction
-    }
+    const int dim_not_zero = iface / 2;//reference direction of face integer division
+//    int dim_not_zero = 0;
+//    for(int idim=0; idim<dim; idim++){
+//        if(std::abs(unit_ref_normal_int[idim]) >= 1e-11)//reference unit normal is 0 in all direction and 1 in outward normal direction
+//            dim_not_zero = idim;//this is outward pointing normal reference direction
+//    }
```
jbrillon, line 1492:
```suggestion
```

src/dg/dg.cpp
```diff
@@ -1443,6 +1468,9 @@ void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, cons
         // assembles and solves for auxiliary variable if necessary.
         assemble_auxiliary_residual();
 
+        dealii::Timer timer;
```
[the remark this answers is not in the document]
AlexanderCicchino: I chose to use Timer without the mpi that way we can compare all of the local CPU times, and sum across processes at the end.

src/dg/strong_dg.cpp
```diff
@@ -1443,11 +1757,36 @@ void DGStrong<dim,nstate,real,MeshType>::assemble_boundary_term_strong(
...
+        }
+       // soln_basis.inner_product_surface_1D(iface, conv_flux_dot_normal_diff[istate], 
+        //Convective surface nnumerical flux.
+        soln_basis.inner_product_surface_1D(iface, conv_flux_dot_normal[istate], 
                                             face_quad_weights, rhs, 
                                             soln_basis.oneD_surf_operator, 
                                             soln_basis.oneD_vol_operator,
-                                            false, -1.0);//adding=false, scaled by factor=-1.0 bc subtract it
+                                            true, -1.0);//adding=true, scaled by factor=-1.0 bc subtract it
+        //Dissipative surface nnumerical flux.
```
jbrillon, line 1789:
```suggestion
        //Dissipative surface numerical flux.
```

src/dg/strong_dg.cpp
```diff
@@ -1443,11 +1757,36 @@ void DGStrong<dim,nstate,real,MeshType>::assemble_boundary_term_strong(
...
+                                            true, -1.0);
+        }
+        else{
+            soln_basis.inner_product_surface_1D(iface, conv_int_vol_ref_flux_interp_to_face_dot_ref_normal[istate], 
+                                                face_quad_weights, rhs, 
+                                                soln_basis.oneD_surf_operator, 
+                                                soln_basis.oneD_vol_operator,
+                                                false, 1.0);//adding=false, scaled by factor=-1.0 bc subtract it
+        }
+       // soln_basis.inner_product_surface_1D(iface, conv_flux_dot_normal_diff[istate],
```
jbrillon, line 1782:
```suggestion
```

src/dg/strong_dg.cpp
AlexanderCicchino: It evaluates ocnv_phys_flux only if it's conservative DG strong, otherwise, if split form or curvilinear split form, it solves the entropy conserving 2pt flux form. Ideally, in the future the EC two-point flux would be its own type and DG strong would only be conservative, but for now I left if statements to differentiate between the two.

src/dg/strong_dg.cpp
```diff
@@ -1740,25 +2093,11 @@ void DGStrong<dim,nstate,real,MeshType>::assemble_face_term_strong(
...
-    const int dim_not_zero = dim_not_zero_probe;
-    if(dim_not_zero == 1000){
-        pcout<<"Error with normals. Normal direction is not defined. Aborting..."<<std::endl;
-        std::abort();
-    }
+    const int dim_not_zero = iface / 2;//reference direction of face integer division
+//    if(dim_not_zero == 1000){
+//        pcout<<"Error with normals. Normal direction is not defined. Aborting..."<<std::endl;
+//        std::abort();
+//    }
```
jbrillon, line 2095: do we still need the check?

AlexanderCicchino: No because I found a simpler way to extract the dim_not_zero directly.

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -421,7 +441,7 @@ void PeriodicTurbulence<dim, nstate>::compute_and_update_integrated_quantities(D
         if (!cell->is_locally_owned()) continue;
         cell->get_dof_indices (dofs_indices);
 
-        // We first need to extract the mapping support points (grid nodes) from high_order_grid.
+// We first need to extract the mapping support points (grid nodes) from high_order_grid.
```
jbrillon, line 444:
```suggestion
        // We first need to extract the mapping support points (grid nodes) from high_order_grid.
```

src/dg/dg.cpp
AlexanderCicchino: Fixed

tests/integration_tests_control_files/advection_explicit_periodic/2D_advection_explicit_periodic_energy_weight_adjusted.prm
```diff
@@ -7,7 +7,7 @@ set dimension = 2
 
 set use_weak_form = false
 
-set overintegration = 0
+set overintegration = 2
```
jbrillon, line 10: why is OI added?

AlexanderCicchino: To check the overintegrated case. Before this PR the test would fail for overintegrated, now it passes.

src/testing/TGV_scaling.cpp
AlexanderCicchino: Yes added the mpi barrier.

src/dg/dg.cpp
```diff
@@ -99,7 +99,7 @@ DGBase<dim,real,MeshType>::DGBase(
     , oneD_quadrature_collection(std::get<7>(collection_tuple))
     , oneD_face_quadrature(max_degree)
     , dof_handler(*triangulation, true)
-    , high_order_grid(std::make_shared<HighOrderGrid<dim,real,MeshType>>(grid_degree_input, triangulation, all_parameters->output_high_order_grid))
+    , high_order_grid(std::make_shared<HighOrderGrid<dim,real,MeshType>>(grid_degree_input, triangulation, all_parameters->check_valid_metric_Jacobian, all_parameters->renumber_dof_handler_Cuthill_Mckee, all_parameters->output_high_order_grid))
```
jbrillon, line 102: use `do_renumber_dofs` and add a function argument `dof_renumbering_type` which is an enum; this would replace the current argument `renumber_dof_handler_Cuthill_Mckee` and allow for easy extension if we ever want to use different renumberings; can open an issue for this or resolve now

jbrillon (commented):
LGTM, left a few comments, great to see the performance boost for our explicit time advancement runs using strong DG / NSFR schemes, thanks for this

src/dg/dg.cpp
```diff
@@ -2143,13 +2179,18 @@ void DGBase<dim,real,MeshType>::allocate_system (
...
+
+    allocate_dual_vector();
 
     // Set use_auxiliary_eq flag
     set_use_auxiliary_eq();
 
     // Allocate for auxiliary equation only.
-    allocate_auxiliary_equation ();
+    if(use_auxiliary_eq)
+        allocate_auxiliary_equation ();
```
jbrillon, line 2190:
```suggestion
    if(use_auxiliary_eq) allocate_auxiliary_equation ();
```

src/dg/dg.cpp
AlexanderCicchino: https://github.com/dougshidong/PHiLiP/issues/220

AlexanderCicchino:
For the overintegration curvilinear tests, it runs out of memory on the p=20 solve after doing the scaling. The last major memory issue for curvilinear overintegration comes from creating the MappingQGeneric object and fe value for get_position_vector in high_order_mesh.  An issue after this PR should be raised to fix this by having an option to extract the volume_nodes vector from the mesh->grids->warp function. This will prevent high_order_mesh from growing in memory proportional to quadrature nodes, but instead being solely dependent on the mesh nodes. Lastly, another large memory usage comes from allocation the dim volume_quadrature_collection and fe_collection_lagrange. For DG strong, neither of these are needed so an option will need to be added where the volume quadrature weights are obtained through the tensor product of the oneD quadrature weights-- similarly for the surface quadrature weights.

AlexanderCicchino:
> 

https://github.com/dougshidong/PHiLiP/issues/221

jbrillon:
I'd also like to acknowledge the size of this PR in comparison to the last one, thanks for the relatively small PR, makes the review much easier for us

src/dg/dg.cpp
```diff
@@ -2095,7 +2129,9 @@ void DGBase<dim,real,MeshType>::allocate_system (
 
     dof_handler.distribute_dofs(fe_collection);
     //This Cuthill_McKee renumbering for dof_handlr uses a lot of memory in 3D, is there another way?
-    if(all_parameters->do_renumber_dofs) dealii::DoFRenumbering::Cuthill_McKee(dof_handler,true);
+    if(all_parameters->renumber_dof_handler_Cuthill_Mckee && all_parameters->do_renumber_dofs){
```
PranshulThakur, line 2132: Just to confirm, `all_parameters->do_renumber_dofs` renumbers solution dofs while `all_parameters->renumber_dof_handler_Cuthill_Mckee` renumbers dofs of both solution and grid coeffs? I guess we currently cannot renumber just the solution dofs (we can probably implement it in future if required).

PranshulThakur (commented):
Thanks for improving scalability of strong DG. I just have one quick comment.

AlexanderCicchino:
[ctest_log_Narval_18May.log](https://github.com/dougshidong/PHiLiP/files/11508589/ctest_log_Narval_18May.log)
The test ANISOTROPIC_MESH_ADAPTATION_SSHOCK failed because I used the wrong version of gmsh. It passes when using gmsh 4.11. 
2D_SUM_FACTORIZATION_HADAMARD_TEST passes locally
MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION passes locally 
MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CPLUS passes locally 
-- they failed on Narval due the cpu usage issue so the scaling is not exact at each delta(poly_degree)
MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CURV and MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CPLUS_CURV should fail at the memory check due to issue https://github.com/dougshidong/PHiLiP/issues/221

src/dg/dg.cpp
AlexanderCicchino: Yes. In future should have dof_handler specific parameters where can have the flags do_renumber_dofs and do_renumber_dofs_grid, then choose specifically what method to renumber the dofs with.

jbrillon:
can you post a log file showing that these pass?

src/dg/dg.cpp
cpethrick: Could you add comments to `all_parameters` indicating the difference between these two parameters as they currently are implemented? They are named so similarly that it could be a source of confusion!

AlexanderCicchino:
> can you post a log file showing that these pass? its helpful that we have an idea of the run time for the tests
[extra_ctest_results.log](https://github.com/dougshidong/PHiLiP/files/11509211/extra_ctest_results.log)

jbrillon:
Thanks, looks good, just add MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CURV and MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CPLUS_CURV as new issues with `testfail`

jbrillon (approved):
LGTM, just address the comment above regarding the 2 failing tests

AlexanderCicchino:
> Thanks, looks good, just add MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CURV and MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CPLUS_CURV as new issues with `testfail`

https://github.com/dougshidong/PHiLiP/issues/223

src/dg/dg.cpp
jbrillon: This will be addressed by issue #220

jbrillon: you could even just add the enum and set `renumber_dof_handler_Cuthill_Mckee` in `al_parameters.cpp` based on `do_renumber_dofs` and `renumber_type`

src/dg/dg.h
```diff
@@ -639,6 +645,10 @@ class DGBase
     /// Sets the current time within DG to be used for unsteady source terms.
     void set_current_time(const real current_time_input);
 
+    /// Computational time for assembling residual.
+   // clock_t assemble_residual_time;
```
cpethrick, line 649:
```suggestion
```

src/mesh/high_order_grid.cpp
```diff
@@ -98,25 +102,33 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::initialize_with
...
+        //Also, when the metric terms are built from operators: metric_operators
+        //it check the validity of the metric Jacobian on-the-fly.
+        const unsigned int exact_jacobian_order = (max_degree-1) * dim;
+        const unsigned int min_jacobian_order = 1;
+        const unsigned int used_jacobian_order = std::max(exact_jacobian_order, min_jacobian_order);
+        evaluate_lagrange_to_bernstein_operator(used_jacobian_order);
+         
+        auto cell = dof_handler_grid.begin_active();
+        auto endcell = dof_handler_grid.end();
+        // pcout << "Disabled check_valid_cells. Took too much time due to shape_grad()." << std::endl;
```
cpethrick, line 120:
```suggestion
```

src/mesh/high_order_grid.cpp
cpethrick, line 129:
```suggestion
```

src/mesh/high_order_grid.cpp
```diff
@@ -98,25 +102,33 @@ void HighOrderGrid<dim,real,MeshType,VectorType,DoFHandlerType>::initialize_with
...
+        for (; cell!=endcell; ++cell) {
+            if (!cell->is_locally_owned())  continue;
+            const bool is_invalid_cell = check_valid_cell(cell);
+         
+            if ( !is_invalid_cell ) {
+                std::cout << " Poly: " << max_degree
+                          << " Grid: " << nth_refinement
+                          << " Cell: " << cell->active_cell_index() << " has an invalid Jacobian." << std::endl;
+                // bool fixed_invalid_cell = fix_invalid_cell(cell);
+                // if (fixed_invalid_cell) std::cout << "Fixed it." << std::endl;
```
cpethrick, line 130:
```suggestion
```

src/operators/operators.cpp
```diff
@@ -809,6 +810,258 @@ void SumFactorizedOperators<dim,n_faces>::Hadamard_product(
...
+void SumFactorizedOperators<dim,n_faces>::sum_factorized_Hadamard_surface_sparsity_pattern(
+    const unsigned int rows_size,
+    const unsigned int columns_size,
+    std::vector<unsigned int> &rows,
+    std::vector<unsigned int> &columns,
+    const int dim_not_zero)
+{
+    //Note that for all directions, the rows vector should always be the same.
+//    const unsigned int n_rows = rows_size;
+//    const unsigned int n_col = columns_size;
```
cpethrick, line 941:
```suggestion
```
Two lines here

src/mesh/high_order_grid.cpp
AlexanderCicchino: This line isn't from this PR.

src/mesh/high_order_grid.cpp
AlexanderCicchino: This line isn't from this PR.

cpethrick (approved):
Looks good to me, thanks for improving the performance of DG strong! I have left a few suggestions where commented code has been left in to make sure that everything is tidy before merging.

Great work!

src/mesh/high_order_grid.cpp
cpethrick: Understood, resolving the ones you've indicated aren't from this PR.

src/operators/operators.cpp
AlexanderCicchino: Fixed.

src/dg/dg.cpp
```diff
@@ -1474,16 +1502,22 @@ void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, cons
...
+                soln_basis_projection_oper_int,
+                soln_basis_projection_oper_ext,
                 mapping_basis,
                 false,
                 right_hand_side,
                 auxiliary_right_hand_side);
         } // end of cell loop
+
+        timer.stop();
+        assemble_residual_time += timer.cpu_time();
```
sivanadarajah, line 1514: Should we consider adding a parameter to check for the time only if we are doing timing runs. If its a production run, wouldn't we be introducing a mpi_waitall to capture the timing?

AlexanderCicchino: Yes we can add a parameter. 
From what I gathered timer.cpu_time() stores the local cpu time for the processor because it was not initialized with the mpi_communicator (line 1471). It does not wait for the other processors to reach the point to store the time. The mpi sum happens in the test after the entire run and that's when the mpi_waitall would be called. If it was initialized like: dealii::Timer timer(this->mpi_communicator,false) then yes it would have an mpi_waitall within it when cpu_time() get called.

src/dg/dg.cpp
AlexanderCicchino: Fixed

PranshulThakur (approved):
LGTM!

PR #227, opened 2023-06-06
Maximum-Principle-Satisfying Limiter

ssaishruthi98:
- Implemented apply_maximum_principle_limiter function in dg that applies a limiter to the global solution using Zhang,Shu May 2010 Eq 3.8 and 3.9
- Implemented tests cases for verification: 1D & 2D Burgers with OOA, 1D Linear Advection, and 2D Linear Advection with OOA


src/dg/dg.cpp
```diff
@@ -3329,7 +3329,206 @@ real2 DGBase<dim,real,MeshType>::discontinuity_sensor(
...
+            for (unsigned int istate = 0; istate < nstate; ++istate) {
+                this->global_max.push_back(-1e9);
+                this->global_min.push_back(1e9);
+            }
+        }
+
+        //get solution coeff
+        std::array<std::vector<real>, nstate> soln_dofs;
+        const unsigned int n_shape_fns = n_dofs_curr_cell / nstate;
+        for (unsigned int idof = 0; idof < n_dofs_curr_cell; ++idof) {
```
sivanadarajah, line 3361: Couldn't we merge this loop with the loop that starts on line 3369?

src/dg/dg.cpp
```diff
@@ -3329,7 +3329,206 @@ real2 DGBase<dim,real,MeshType>::discontinuity_sensor(
...
+    soln_basis_projection_oper.build_1D_volume_operator(this->oneD_fe_collection_1state[this->max_degree], this->oneD_quadrature_collection[this->max_degree]);
 
+
+    if (this->global_max.empty() && this->global_min.empty())
+    {
+        this->get_global_max_and_min_of_solution();
+    }
+
+    auto metric_cell = this->high_order_grid->dof_handler_grid.begin_active();
+    for (auto soln_cell = this->dof_handler.begin_active(); soln_cell != this->dof_handler.end(); ++soln_cell, ++metric_cell) {
```
sivanadarajah, line 3410: Why not use active_cell_iterators?

PR #235, opened 2023-08-03
Serrated Edges 3D Potential Flow Simulation

aidanchristoffersson:
Rough draft for review.


src/dg/dg.cpp
```diff
@@ -371,6 +366,10 @@ bool DGBaseState<dim,nstate,real,MeshType>::potential_body_geometry(
 
 
         // checking if position within TES geometry
+        // TES shape
+        auto TES = [&](real v) {
```
jbrillon, line 370: just wondering, what does `[&](real v)` do? I've never seen this before

aidanchristoffersson: Passes the real variable v as a reference. (real v) is just a stand in for whatever variable is passed  in the expression.

See the below article for reference, my understanding is lambdas are just a replacement for function pointers. In this case not really needed now that I look back on it... Could just be written inline.

https://en.cppreference.com/w/cpp/language/lambda

PR #242, opened 2023-09-27
Finite wing adaptive sampling

alex-fricker:
Introduction of a test for the adaptive sampling algorithm for a 3D naca 0012 wing with finite span.


alex-fricker:
This was a mistake - please ignore.

PR #243, opened 2023-09-29
Clean up explicit and default construct destructor

dougshidong:
Remove unnecessary empty constructors (those exist by default). 

Remove unnecessary empty and destructors (those exist by default). https://sonarcloud.io/organizations/dougshidong/rules?open=cpp%3AS4963&rule_key=cpp%3AS4963

The abstract classes which have a virtual destructor must have an implementation (usually ` = default`). https://sonarcloud.io/organizations/dougshidong/rules?languages=cpp&open=cpp%3AS1235&q=destructor+virtu

Use `default` in case the constructor or destructor is absolutely necessary instead of `Foo::Foo() {};`

Declare constructors with `explicit` keyword. https://sonarcloud.io/organizations/dougshidong/rules?open=cpp%3AS1709&rule_key=cpp%3AS1709


dougshidong:
For some reason the diff shows commits previously merged in https://github.com/dougshidong/PHiLiP/pull/241. The diff is really just the last 2 commits. The tab Files Changed is accurate.

src/grid_refinement/msh_out.h
```diff
@@ -34,10 +34,13 @@ template <int dim>
 class MshOutData
 {
 public:
+    /// Destructor
+    virtual ~ MshOutData() = default;
```
cpethrick, line 38:
```suggestion
    virtual ~MshOutData() = default;
```

src/optimization/flow_constraints.hpp
```diff
@@ -94,6 +94,7 @@ class FlowConstraints : public ROL::Constraint_SimOpt<double> {
     //    const FreeFormDeformation<dim> &_ffd,
     //    std::vector< std::pair< unsigned int, unsigned int > > &_ffd_design_variables_indices_dim,
     //    const dealii::TrilinosWrappers::SparseMatrix *existing_dXvdXp = NULL);
```
cpethrick, line 96: Remove commented constructor?

cpethrick (commented):
Thanks for another cleanup PR! I don't have many comments since the changes are all very similar.

There are several tests which take only a parameters pointer as input to the constructor. Should those also use `explicit`?

dougshidong:
> Thanks for another cleanup PR! I don't have many comments since the changes are all very similar.
> 
> There are several tests which take only a parameters pointer as input to the constructor. Should those also use `explicit`?

Good catch. I have SonarLint detecting all the errors I had fixed locally. However, it's not including the `testing` directory in the global analysis, only when I'm running it on the specific file. Looking into why all the testing files have been missed by the local analysis. The web analysis seems to have caught them all.

src/operators/operators.h
```diff
@@ -53,6 +53,9 @@ template <int dim, int n_faces>
 class OperatorsBase
 {
 public:
+    /// Destructor
+        virtual ~OperatorsBase() = default;
```
jbrillon, line 57:
```suggestion
    virtual ~OperatorsBase() = default;
```

jbrillon (approved):
Great PR, thanks for another clean up!

dougshidong:
Alright, found all them. No idea why my local SonarLint exclude files from analysis. Asked in their [forums](https://community.sonarsource.com/t/sonarlint-for-intellij-clion-header-files-excluded-compiler-kind-not-found/101843). @cpethrick , could you review/approve?

cpethrick (approved):
Looks good! Thanks for fixing the rest of the testing directory.

PR #245, opened 2023-10-05
Strong DG NSFR boundary integral now uses projected entropy variables.

AlexanderCicchino:
I read through the strong DG boundary integral, saw that the surafce numerical flux used the solution interpolated to the surface, and updated the convective surface numerical flux and boundary term to use the projected entropy variables. There is no test verifying this currently. A good test would be checking entropy conservation with the boundary condition entropy production to machine precision.


src/dg/strong_dg.cpp
```diff
@@ -1719,19 +1734,23 @@ void DGStrong<dim,nstate,real,MeshType>::assemble_boundary_term_strong(
...
+        //I am not sure if BC should be from solution interpolated to face
+        //or solution from the projected entropy variables.
+        //Now, it uses projected entropy variables for NSFR, and solution
+        //interpolated to face for conservative DG.
+        this->pde_physics_double->boundary_face_values (boundary_id, surf_flux_node, unit_phys_normal_int, soln_state_int, aux_soln_state_int, soln_boundary, grad_soln_boundary);
         
         // Convective numerical flux.
         std::array<real,nstate> conv_num_flux_dot_n_at_q;
-        conv_num_flux_dot_n_at_q = this->conv_num_flux_double->evaluate_flux(soln_state, soln_boundary, unit_phys_normal_int);
+        conv_num_flux_dot_n_at_q = this->conv_num_flux_double->evaluate_flux(soln_state_int, soln_boundary, unit_phys_normal_int);
```
sivanadarajah, line 1745: @AlexanderCicchino This is a good catch but we need to do a more thorough review (with another student) including the boundary terms since soln_boundary and the gradient still depend on _int values.

dougshidong: > A good test would be checking entropy conservation with the boundary condition entropy production to machine precision.

Would the aforementionned test have caught this?

AlexanderCicchino:
> That would be a great test to have along with this PR while you're fixing it. The odds that someone puts a test for this in 4 years are pretty much nil and currently open issues don't exactly get resolved.
> 
> Approving since a fix is better than no fix, but highly suggesting a test.
> 
> Also, at this point, split form should pretty much be its own form of DG.

I agree. I didn't put a test because I was under the impression that another student is looking into applying a test using BC with NSFR.
Yes I also agree that the split form should pretty much be its own form from DG by this point.

src/dg/strong_dg.cpp
AlexanderCicchino: Yes, I agree Prof. Nadarajah. 
Yes Doug, the aforementioned test would catch it, provided it checks both entropy conservation and global conservation.

jbrillon (approved):
Thanks for catching this @AlexanderCicchino and for the nice PR!

Issue #249, opened 2023-11-23
Strong DG is producing nonphysical results with NACA airfoil test case


cpethrick:
I've been running into issues with strong DG on grids which are more complicated than just a periodic cube. To illustrate this as a minimal working example, I've reproduced the issue using a NACA0012 airfoil which is generated using deal.ii (to eliminate any uncertainty about GMSH). Please see a [branch](https://github.com/cpethrick/PHiLiP/tree/strong_vs_weak_dealii_naca_mesh) on my repository to run these; I have minimally changed the subsonic NACA0012 case to be unsteady and use RK4 such that weak and strong can be compared.

Using weak DG, the result after a small time interval looks reasonable:
![image](https://github.com/dougshidong/PHiLiP/assets/94074298/59ff612c-c5a2-4e21-83a1-1665736b2808)

But strong DG results in strange jumps, which are aligned with cell faces:
![image](https://github.com/dougshidong/PHiLiP/assets/94074298/31c95a2e-2ab4-44ad-abd4-ec3dd85c3b96)

These discontinuities appear at the very first time step. I have seen this on a few GMSH meshes, as has Shruthi, and the commonalities seem to be that we have non-periodic BCs and interfaces between "zones" of the mesh where cell shapes or densities change. The issue seems to go away when grids are perfectly cartesian.

I'm happy to hear any input from others, as this could caused by any number of things, including BCs, the face integral, mesh generation, etc.

Issue #240, opened 2023-09-20
Gradient at boundary not enforced for viscous flows


jbrillon:
Currently, the gradient at the physical boundary (`soln_grad_ext`) set by `boundary_face_values` is not being enforced; as it is being set to the interior value (`soln_grad_int`) in all cases inside `evaluate_auxiliary_flux`.

To fix it, we must pass `on_boundary` flag to PhysicsBase's `dissipative_flux` and modify the dissipative flux accordingly; the fix take effect in `assemble_boundary_term`

jbrillon:
will be fixed by https://github.com/jbrillon/PHiLiP/pull/11 once PR #199 goes through

PR #250, opened 2023-11-25
Fixed exterior reference normal zone issue in strong. 

AlexanderCicchino:
This PR fixes/closes the zone issue in strong DG as per issue #249 . The problem was that previously, on a face, it was assumed that the reference faces were neighbours, which is the case for grids made from hypercube and a warping or a single mesh in dealii. In the general case, this isn't, and now it has been fixed. Note, the reference cell itself could be rotated and no longer be a square going from 0->1, that case has not been tested in strong.
strong_naca0012_solution


cpethrick:
Thanks for fixing this, Alex! I'm really glad that this was able to be resolved so quickly.

I think it's good to add these new tests, but we should have a testfail condition. I can do that in the next few days: it will check that weak and strong have the expected lift (with some small tolerance) after the short unsteady time interval. I'll add it to this branch once it's done.

AlexanderCicchino:
My pleasure, hope it helps!

dougshidong:
> Thanks for fixing this, Alex! I'm really glad that this was able to be resolved so quickly.
> 
> I think it's good to add these new tests, but we should have a testfail condition. I can do that in the next few days: it will check that weak and strong have the expected lift (with some small tolerance) after the short unsteady time interval. I'll add it to this branch once it's done.

+1 to the test success / failure condition.

jbrillon (approved):
LGTM, @cpethrick could you run ctest and post the results once you've added the lift coefficient check test case?

cpethrick:
I just added a new test that checks lift and drag for the three NACA0012 unsteady tests (weak/strong/strong-split). They currently pass on this branch, but the two strong versions fail on master.

I am starting CTEST. It should finish some time tomorrow.

src/testing/naca0012_unsteady_check_quick.cpp
```diff
@@ -0,0 +1,75 @@
...
+        drag_expected = 0.220394;
+    }else if (this->all_parameters->use_split_form) {
+        // Strong, Ra two-point flux
+        lift_expected = 0.24326;
+        drag_expected = 0.268897;
+    } else {
+        // Strong, Roe flux
+        lift_expected = 0.198245;
+        drag_expected = 0.220394;
+    }
```
dougshidong, line 39: If we're going to use magic numbers that depend on the number of steps, step size, final time, numerical flux, etc. they should also be hardcoded here. That way, if someone changes one thing, such as the number of steps, they see the surrounding magic numbers that would also be affected. Therefore, the above should specify any non-default parameters. The parameter file should contain variations for which this test passes, (e.g. weak form, strong form, strong Ra two-point flux).

I also see that both strong and weak use the same expected changes. If they are expected to be the same, let's use 1 set of magic numbers. Also, what makes us believe that the `Strong, Ra two-point flux` is correct? If `Strong, Ra two-point flux` starts failing because of a different lift without any obvious change, how do we know that the change was incorrect or that the old version was incorrect?

Alternatively to all the above comments, I think that a more robust test here would be to run both strong and weak DG, and then compare the lift and drag, or any other expected observable measure. That way, the test doesn't depend on magic values and works for any parameter input. The intent of the test is therefore "whatever weak DG does, strong DG does similarly as well", if nothing else breaks, one of the two is wrong.

cpethrick: Responding to each paragraph:
1. I structured the test similarly to the TGV energy tests, which specify an expected energy in the prm file ([e.g](https://github.com/dougshidong/PHiLiP/blob/a29ecce53f33a3f0bafc14ef8db767fe967e29b8/tests/integration_tests_control_files/taylor_green_vortex_integration/viscous_taylor_green_vortex_energy_check_strong_quick.prm#L66) ). I think the approach in those tests is a bit better because the expected energy is in the same .prm file as all other parameters. However, since I don't see this test being re-used extensively and I wanted to avoid adding more test-specific parameters. I recognize that this isn't ideal in terms of coding practice, and I can change it if needed.
2. At this time step size, strong and weak yield the same lift and drag, which I believe we should expect for a subsonic flow per the discussion on discord. I can change it to be a single condition, checking `if ( (strong || weak) && not split) `.  I didn't do this initially as when the time step was bigger, they had different values. Re: "What makes us believe that `Strong, Ra two-point flux` is correct?", I anticipate that the current values may change with Julien's changes to adiabatic wall boundaries -- see issue #240. The test will fail if there are changes which impact lift or drag. We won't know which version is correct without other tests, but this test will alert us to any major changes; if there is reason to believe a new version is correct, the hard-coded "magic numbers" will need to be changed. The console output should make it easy to recognize why a test is failing.
3. I agree that this would be a more robust way to compare weak and strong DG if the user is expected to be changing the parameters file. However, I don't think we can expect that split form strong DG will yield the same lift and drag as non-split DG, as the flux is different. Therefore, we would still need this sort of "magic number" test for split form strong, as there is no weak equivalent. 

I can come back to this PR either tomorrow or Friday. Of the following options, which would you prefer?
1. Keep the test as-is, with hard-coded lift and drag values. Remove the duplicate lift and drag such that there are only two sets.
2. Move the expected lift and drag into parameters, consistent with the TGV energy check tests. Having them in the same place as the test parameters will reduce the likelihood that a user changes a parameter and gets an unexpected testfail. However, this will mean more test-specific parameters.
3. Make an additional test which compares the lift and drag resulting from strong and weak DG with otherwise identical parameters. There will still be a hard-coded strong-split test.

cpethrick: Update: I just pushed a new version of the test. It compares strong and weak with Roe flux, using a single test and no hard-coded numbers. There is no check for strong-split at this time, but I have a framework set up for it in the test.

dougshidong (approved):

Issue #223, opened 2023-05-18
MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CURV and MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CPLUS_CURV fail from memory


AlexanderCicchino:
MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CPLUS_CURV and MPI_3D_TAYLOR_GREEN_SCALING_OVERINTEGRATION_CURV fail at the p=20 memory allocation due to the issue https://github.com/dougshidong/PHiLiP/issues/221

tofstie:
I am closing this issue as these tests should no longer fail from memory. Issue #221 still stays as the issues relating to why the memory is being allocated has not yet been fixed.

PR #253, opened 2024-01-19
Templated Operators with typename real.

AlexanderCicchino:
Templated operators class with type real for input, output, and weight vectors. The basis is assumed to stay as double for the sum-factorization. This allows operators to be used with AD type.


AlexanderCicchino:
Currently running ctest on Narval. Will post log file once it is done.

AlexanderCicchino:
[output_ctest_Jan_25_2024.log](https://github.com/dougshidong/PHiLiP/files/14067492/output_ctest_Jan_25_2024.log)

AlexanderCicchino:
Fixed [Issue](https://github.com/dougshidong/PHiLiP/issues/223).

cpethrick (commented):
Thanks for making this change! I have confirmed that everything builds. As there are no changes to the code's current functionality, should we leave a note somewhere indicating that these templating changes are untested?

My one comment on this PR is that the alignment of some functions has been impacted (e.g., line 674 of dg_base.hpp). If it doesn't take too long, could you restore the alignment on the dg classes?

Otherwise, thank you for going through the effort to template the operators classes!

AlexanderCicchino:
Yes of course! Thanks for the catch! Fixed the alignment.

cpethrick (approved):

PranshulThakur (commented):
Thanks a lot @AlexanderCicchino for templating Operators! It will help significantly to compute sensitivities with strong DG in future research projects (and probably weak DG too if we update it to use sum factorization). The PR looks great! 

To use strong DG with AD, do we need to make any additional modifications to strong DG? Also, is it possible to test AD on strong DG for inviscid cases by making slight modifications to the dRdW and dRdX unit tests in tests/unit_tests/sensitivities/?

AlexanderCicchino:
Thanks @PranshulThakur! To use strong DG with AD: for inviscid problems it will need the modifications from weak DG's assemble_volume_term_derivatives for the dRdW and dRdX. I don't think there needs to be any changed in assemble volume/face/boundary functions. For viscous problems, strong DG's DR/Du will need to incorporate the auxiliary variable q and its respective solution:  DR/Du +DR/Dq*Dq/Du. Again, this can be computed outside of the assemble volume/face/boundary functions for the primary and auxiliary equations.

PranshulThakur:
> Thanks @PranshulThakur! To use strong DG with AD: for inviscid problems it will need the modifications from weak DG's assemble_volume_term_derivatives for the dRdW and dRdX. I don't think there needs to be any changed in assemble volume/face/boundary functions. For viscous problems, strong DG's DR/Du will need to incorporate the auxiliary variable q and its respective solution: DR/Du +DR/Dq*Dq/Du. Again, this can be computed outside of the assemble volume/face/boundary functions for the primary and auxiliary equations.

Makes sense. Thanks for the explanation.

PranshulThakur (approved):
LGTM!

PR #255, opened 2024-03-20
Fix for output files being overwritten

ssaishruthi98:
When outputting files using the parameter `output_solution_every_dt_time_intervals` there are instances where casting to int results in the same `file_number` being used twice. When the same `file_number` is used twice, the previous output file is overwritten with the newer output. Examples of these cases are shown below.

fileOutputBug_1
fileOutputBug_2

To fix this I changed the type for `file_number` to double so that the round function can be used. With this fix, output files are no longer being overwritten.


cpethrick (approved):
Thanks for adding this fix!

Per our discussion, the changes look good, and it builds in both release and debug. Personally, I would prefer that you explicitly typecast (i.e., `const int file_number = (int)(round(...))`), but I will leave that up to you. That's my only comment, therefore I've marked this as approved from my end.

In my opinion, this is a minor change so it's okay to merge without running ctest.

jbrillon (approved):
LGTM

PR #259, opened 2024-06-17
Low Storage Runge-Kutta

jocelynpoon:
### PR for the Low Storage RK Method
Code for [3S*] and [3S*+] methods. Tableaus for RK4(3)5[3S*],  RK3(2)5F[3S*+], RK4(3)9F[3S*+], and RK4(4)10F[3S*+] schemes. 

### low_storage_runge_kutta_ode_solver.cpp

- Algorithms for [3S*] and [3S*+] methods
- Error adaptive step size
- Automatic initial step size

### low_storage_rk_tableau_base.cpp

- Sets tableau and gets tableau values

### low_storage_runge_kutta_methods.cpp

- Tableaus for RK4(3)5[3S*],  RK3(2)5F[3S*+], RK4(3)9F[3S*+], and RK5(4)10F[3S*+] schemes. 
- Tableaus include gamma, beta, delta, b^hat coefficients

### parameters_ode_solver.cpp

- Add methods to ODE solver parameters
- Add subsection low-storage rk solver to prm file to set atol, rtol, beta1, beta2, beta3

ctest log: (some tests that failed in the log don't fail anymore because I reverted the tests back to what's on master)
[log.txt](https://github.com/user-attachments/files/16628666/log.txt)


src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,203 @@
...
+    //u_hat = s2;
+
+    dealii::LinearAlgebra::distributed::Vector<double> rhs = s1;
+
+    int m = 5;
+    //int q_hat = 3;
+    //int k = q_hat +1;
+    
+    double sum_delta = 0;
+    double atol = 0.1;
```
sivanadarajah, line 51: Lets include atol and rtol as parameters

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+template <int dim, typename real, int n_rk_stages, typename MeshType> 
+double LowStorageRungeKuttaODESolver<dim,real,n_rk_stages,MeshType>::err_time_step (real dt, const bool pseudotime)
+{
+    (void) pseudotime;
+    int q_hat = 3;
+    int k = q_hat +1;
+   // double error = 0.0;
+    //double w = 0.0;
+    double beta_controller[3] = {0.70, -0.40, 0};
```
sivanadarajah, line 132: The same for beta as well.

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    //double atol = 0.001;
+    //double rtol = 0.001;
+    double error = 0.0;
+    w = 0.0;
+
+    for (int i = 1; i < n_rk_stages +1; i++ ){
+        // storage_register_2 = storage_register_2 + delta[i-1] * storage_register_1;
+        storage_register_2.add(this->butcher_tableau->get_delta(i-1) , storage_register_1);
+        this->dg->solution = rhs;
```
cpethrick, line 52: I think this might be possible to do without the extra vector `rhs`, and instead using `storage_register_1`. I haven't gone through the loop in detail, but maybe you could check that.

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,188 @@
...
+        this->dg->assemble_residual();
+        this->dg->apply_inverse_global_mass_matrix(this->dg->right_hand_side, rhs);
+        rhs *= dt;
+        storage_register_4.add(this->butcher_tableau->get_b_hat(n_rk_stages), rhs);
+        //this->pcout << " b_hat_fsal " << this->butcher_tableau->get_b_hat(n_rk_stages);
+
+    }
+    this->dg->solution = storage_register_1;
+    
+    double global_size = dealii::Utilities::MPI::sum(storage_register_1.local_size(), this->mpi_communicator);
```
cpethrick, line 94: This would probably be better to do in `allocate_ode_system` and storing as a member variable. Summing over all of the cores is often quite slow, especially as you increase the number of parallel processes.

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,188 @@
...
+    this->solution_update = this->dg->solution; //storing u_n
+    (void) pseudotime;
+
+    dealii::LinearAlgebra::distributed::Vector<double> storage_register_2;
+    storage_register_2.reinit(this->solution_update);
+    storage_register_2*=0;
+    dealii::LinearAlgebra::distributed::Vector<double> storage_register_1;
+    storage_register_1.reinit(this->solution_update);
+    storage_register_1 = this->dg->solution;
+    dealii::LinearAlgebra::distributed::Vector<double> storage_register_3 = storage_register_1;
```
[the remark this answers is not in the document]
cpethrick: Just be careful to reinitialize the vectors if that needs to be done!

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,188 @@
...
+        rhs *= dt;
+        storage_register_4.add(this->butcher_tableau->get_b_hat(n_rk_stages), rhs);
+        //this->pcout << " b_hat_fsal " << this->butcher_tableau->get_b_hat(n_rk_stages);
+
+    }
+    this->dg->solution = storage_register_1;
+    
+    double global_size = dealii::Utilities::MPI::sum(storage_register_1.local_size(), this->mpi_communicator);
+    
+    if (this->butcher_tableau->get_b_hat(1) == 0){
```
cpethrick, line 96: It would be easier to read if you made a `const bool is_3Sstarplus` to control this logic (and above, too). 

Also, I think the computation of `w` would be better to do in `err_time_step`. If you make the storage registers member variables, that should be an easy change!

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    ++(this->current_iteration);
+    this->current_time += dt;
+    this->pcout << " Time is: " << this->current_time <<std::endl;
+    this->pcout << "dt" << dt;
+}
+
+
+template <int dim, typename real, int n_rk_stages, typename MeshType> 
+double LowStorageRungeKuttaODESolver<dim,real,n_rk_stages, MeshType>::err_time_step (real dt, const bool pseudotime)
```
cpethrick, line 128: Can we use a name similar to the ones used in flow solver for consistency? Maybe `get_automatic_error_adaptive_step_size`.

src/ode_solver/low_storage_runge_kutta_ode_solver.h
```diff
@@ -0,0 +1,75 @@
...
+    std::vector<bool> butcher_tableau_aii_is_zero;
+/*
+    std::vector<dealii::LinearAlgebra::distributed::Vector<double>> storage_register_2;
+    std::vector<dealii::LinearAlgebra::distributed::Vector<double>> storage_register_1;
+    std::vector<dealii::LinearAlgebra::distributed::Vector<double>> storage_register_3;
+    std::vector<dealii::LinearAlgebra::distributed::Vector<double>> storage_register_4;
+    std::vector<dealii::LinearAlgebra::distributed::Vector<double>> rhs;
+*/ 
+    
+    real w;
```
cpethrick, line 61: Could you add comments for all new variables/functions? The syntax can be copied from above.

src/ode_solver/low_storage_runge_kutta_ode_solver.h
```diff
@@ -0,0 +1,75 @@
...
+#include "JFNK_solver/JFNK_solver.h"
+#include "dg/dg_base.hpp"
+#include "ode_solver_base.h"
+#include "runge_kutta_methods/low_storage_rk_tableau_base.h"
+#include "relaxation_runge_kutta/empty_RRK_base.h"
+
+namespace PHiLiP {
+namespace ODE {
+
+/// Runge-Kutta ODE solver (explicit or implicit) derived from ODESolver.
```
cpethrick, line 13: Could you add the reference to the article we are using? Any reference style is okay, so long as it has year, authors, title, etc. Also add a link to that github page which has the correct coefficients.

src/ode_solver/low_storage_runge_kutta_ode_solver.h
```diff
@@ -0,0 +1,75 @@
...
+
+    /// Stores functions related to relaxation Runge-Kutta (RRK).
+    /// Functions are empty by default.
+    std::shared_ptr<EmptyRRKBase<dim,real,MeshType>> relaxation_runge_kutta;
+
+    /// Implicit solver for diagonally-implicit RK methods, using Jacobian-free Newton-Krylov 
+    /** This is initialized for any RK method, but solution-sized vectors are 
+     *  only initialized if there is an implicit solve
+     */
+    JFNKSolver<dim,real,MeshType> solver;
```
cpethrick, line 46: I think this can be removed since we will never have implicit for this type of low-storage algorithm. You can remove it from the header here and also from line 13 of the cpp file.

src/ode_solver/runge_kutta_methods/low_storage_runge_kutta_methods.h
```diff
@@ -0,0 +1,54 @@
+#ifndef __LOW_STORAGE_RUNGE_KUTTA_METHODS__
+#define __LOW_STORAGE_RUNGE_KUTTA_METHODS__
+
+#include "low_storage_rk_tableau_base.h"
+
+namespace PHiLiP {
+namespace ODE {
+
+#if PHILIP_DIM==1
```
cpethrick, line 9: Could you also add references to the paper and github repo to this class?

src/parameters/parameters_linear_solver.cpp
```diff
@@ -7,6 +7,18 @@ void LinearSolverParam::declare_parameters (dealii::ParameterHandler &prm)
 {
     prm.enter_subsection("linear solver");
     {
+        prm.declare_entry("atol", "0.001",
```
cpethrick, line 10: Could these be moved to the ode_solver parameters section rather than this one? I think they would be more relevant there.

src/parameters/parameters_ode_solver.cpp
```diff
@@ -102,7 +104,9 @@ void ODESolverParam::declare_parameters (dealii::ParameterHandler &prm)
                           " euler_ex | "
                           " euler_im | "
                           " dirk_2_im | "
-                          " dirk_3_im"),
+                          " dirk_3_im | "
+                          " RK3_2_5F_3SStarPlus | "
+                          " RK4_3_5_3SStar "),
                           "Runge-kutta method to use. Methods with _ex are explicit, and with _im are implicit."
```
cpethrick, line 110: Maybe add another line to this description indicating that 2SStarPlus and 3SStar are low-storage RK methods.

src/parameters/parameters_ode_solver.h
```diff
@@ -61,12 +62,15 @@ class ODESolverParam
         euler_ex, ///Forward Euler
         euler_im, ///Implicit Euler
         dirk_2_im, ///Second-order diagonally-implicit RK
-        dirk_3_im ///Third-order diagonally-implicit RK
+        dirk_3_im, ///Third-order diagonally-implicit RK
+        RK4_3_5_3SStar,
```
cpethrick, line 66: Add short comments with the main properties - include order, number of stages, FSAL, and other pertinent information.

cpethrick (commented):
This is great work! I'm glad to see it coming together so well. I've left some preliminary comments to work through while the PR is still in draft. As you clean up your code, it is also good to remove commented lines,  though I know those are sometimes necessary while you're still developing it!

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
jocelynpoon: I was looking into this but I think it would be more complicated using storage_register_1 instead of rhs because in that step to calculate storage_register_1, it also uses the unchanged value of storage_register_1. I looked through the Julia code that the paper references and they also use another variable to store f(storage_register_1)

src/parameters/parameters_ode_solver.cpp
```diff
@@ -7,6 +7,26 @@ void ODESolverParam::declare_parameters (dealii::ParameterHandler &prm)
 {
     prm.enter_subsection("ODE solver");
     {
+        prm.declare_entry("atol", "0.001",
+                          dealii::Patterns::Double(),
+                          "Absolute Tolerance");
```
cpethrick, line 12: Could you add a little more detail to these comments? Maybe add "For automatic step size controller". 

It would be nice to also have these in their own subsection, but that's up to preference! ( see the line with ` prm.enter_subsection("rrk root solver");` for an example).

src/flow_solver/flow_solver.cpp
```diff
@@ -500,19 +500,23 @@ int FlowSolver<dim,nstate>::run() const
...
+            next_time_step = ode_solver->err_time_step(time_step,false); 
+            // pcout << "next time step " << next_time_step;
             // advance solution
-            ode_solver->step_in_time(time_step,false); // pseudotime==false
+            // ode_solver->err_time_step(time_step); // pseudotime==false
 
             // Compute the unsteady quantities, write to the dealii table, and output to file
             flow_solver_case->compute_unsteady_data_and_write_to_table(ode_solver, dg, unsteady_data_table);
             // update next time step
-            if(flow_solver_param.adaptive_time_step == true) {
```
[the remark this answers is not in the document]
cpethrick: Sweet, looks good!

src/flow_solver/flow_solver.cpp
```diff
@@ -443,6 +443,9 @@ int FlowSolver<dim,nstate>::run() const
         if(flow_solver_param.adaptive_time_step == true) {
             pcout << "Setting initial adaptive time step... " << std::flush;
             time_step = flow_solver_case->get_adaptive_time_step_initial(dg);
+        } else if(flow_solver_param.error_adaptive_time_step == true) {
+            pcout << "Setting initial error adaptive time step... " << std::flush;
```
cpethrick, line 447: Nice! The organization looks good.

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
cpethrick: Okay! Thanks for checking their code.

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
cpethrick: Looks good!

src/ode_solver/low_storage_runge_kutta_ode_solver.h
cpethrick: Could you add comments also for the extra storage registers you've added?

src/ode_solver/low_storage_runge_kutta_ode_solver.h
```diff
@@ -0,0 +1,105 @@
...
+#endif
+class LowStorageRungeKuttaODESolver: public ODESolverBase <dim, real, MeshType>
+{
+public:
+    LowStorageRungeKuttaODESolver(std::shared_ptr< DGBase<dim, real, MeshType> > dg_input,
+            std::shared_ptr<LowStorageRKTableauBase<dim,real,MeshType>> rk_tableau_input,
+            std::shared_ptr<EmptyRRKBase<dim,real,MeshType>> RRK_object_input); ///< Constructor.
+
+    /// Function to evaluate automatic error adaptive time step
+    double get_automatic_error_adaptive_step_size(real dt, const bool pseudotime);
```
cpethrick, line 35: I think you should add `override` to these.

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,258 @@
...
+    this->pcout << std::endl;
+
+    // sum over all elements
+    w = dealii::Utilities::MPI::sum(w, this->mpi_communicator);
+    w = pow(w / global_size, 0.5);
+    epsilon[2] = epsilon[1];
+    epsilon[1] = epsilon[0];
+    epsilon[0] = 1.0 / w;
+    dt = pow(epsilon[0], 1.0 * beta1/rk_order) * pow(epsilon[1], 1.0 * beta2/rk_order) * pow(epsilon[2], 1.0 * beta3/rk_order) * dt;
+    this->pcout << " dt: " << dt;
```
cpethrick, line 141: There is a parameter `ode_param.ode_output` which could be used to control whether debugging information is printed. See this [line](https://github.com/dougshidong/PHiLiP/blob/5ec59ea934e6577aa7c71ef64f64602ee8e29ab8/src/ode_solver/implicit_ode_solver.cpp#L31) for example. Could you add a similar structure to control debug printing?

It's completely fine to just use `this->pcout` without checking the output mode if you're still doing personal testing/debugging, but using the `ode_output` param will be helpful when this is in master!

cpethrick (commented):
This is looking really good! Thanks for addressing my comments :) I have a few more small comments to address when you have the time.

src/parameters/parameters_flow_solver.cpp
```diff
@@ -91,6 +91,10 @@ void FlowSolverParam::declare_parameters(dealii::ParameterHandler &prm)
                           dealii::Patterns::Bool(),
                           "Solve steady-state solution. False by default (i.e. unsteady by default).");
 
+        prm.declare_entry("error_adaptive_time_step", "false",
+                          dealii::Patterns::Bool(),
+                          "Adapt the time step on the fly for unsteady flow simulations. False by default (i.e. constant time step by default).");
```
cpethrick, line 96: Could you modify this comment and the one for `adaptive_time_step` to describe how the two are different? e.g., your new method would say "adapt according to an estimate of temporal error" (or words to that effect), and the old one would say "adapt according to a CFL condition".

cpethrick (commented):
Great work! I've left some more comments to clean up the PR. It is looking great already and you should be proud of this work! Let me know if you have questions about any of these, and hopefully they will all be quick to resolve.

src/flow_solver/flow_solver.cpp
```diff
@@ -441,8 +441,13 @@ int FlowSolver<dim,nstate>::run() const
         //----------------------------------------------------
         double time_step = 0.0;
         if(flow_solver_param.adaptive_time_step == true) {
+            pcout << "WARNING: adaptive time step is different from error adaptive time step " << std::endl;
```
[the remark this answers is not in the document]
ssaishruthi98: I think it might be best to abort as suggested. From the code it looks like the test will move forward with the adaptive time step instead of the error adaptive step and if the user misses the warning then the results could be confusing since a constant time step would be used throughout.

src/flow_solver/flow_solver.cpp
```diff
@@ -500,19 +505,23 @@ int FlowSolver<dim,nstate>::run() const
             }
 
             // update time step in flow_solver_case
-            flow_solver_case->set_time_step(time_step);
-
+            ode_solver->step_in_time(time_step,false);
+
```
ssaishruthi98, line 509: If the step_in_time function has been moved ahead of the advance solution comment, can the comment be removed as it doesn't pertain to anything?

src/flow_solver/flow_solver.cpp
```diff
@@ -500,19 +505,23 @@ int FlowSolver<dim,nstate>::run() const
             }
 
             // update time step in flow_solver_case
-            flow_solver_case->set_time_step(time_step);
```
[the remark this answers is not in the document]
ssaishruthi98: If it doesn't impact other cases, we should probably remove the comment before it that says "update time step in flow_solver_case" since that is not being done anymore.

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -329,6 +329,9 @@ void PeriodicTurbulence<dim, nstate>::compute_and_update_integrated_quantities(D
     // Initialize the maximum local wave speed to zero; only used for adaptive time step
     if(this->all_param.flow_solver_param.adaptive_time_step == true) this->maximum_local_wave_speed = 0.0;
 
+    // Initialize the maximum local wave speed to zero; only used for adaptive time step
+    if(this->all_param.flow_solver_param.error_adaptive_time_step == true) this->maximum_local_wave_speed = 0.0;
+
```
ssaishruthi98, line 334: Since this if clause does the same thing as the previous one it might be better to combine the statement.

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+    else if (is_3Sstarplus == true){
+        // loop sums elements at each mpi processor
+        for (dealii::LinearAlgebra::distributed::Vector<double>::size_type i = 0; i < storage_register_1.local_size(); ++i) {
+            error = storage_register_1.local_element(i) - storage_register_4.local_element(i);
+            w = w + pow(error / (atol + rtol * std::max(std::abs(storage_register_1.local_element(i)), std::abs(storage_register_4.local_element(i)))), 2);
+        }
+    }
+    this->pcout << std::endl;
```
ssaishruthi98, line 136: Would this std::endl statement result in unnecessary new lines since there is no other print statements within this function?

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+    d0 = u_n.linfty_norm();
+    
+    this->dg->solution = rhs_initial;
+    this->dg->assemble_residual();
+    this->dg->apply_inverse_global_mass_matrix(this->dg->right_hand_side, rhs_initial);
+
+    d1 = rhs_initial.linfty_norm();
+
+    if (d0 < pow(10, -5) || d1 < pow(10, -5)){
```
ssaishruthi98, line 180: I think you can just do `1e-5` instead of the pow for simplicity.

ssaishruthi98 (commented):
Great job on this PR! It's awesome work 😄 
I've left a few comments and questions, please let me know if I can help clarify any of it. I think it should all be easy fixes though!

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
```diff
@@ -0,0 +1,270 @@
...
+    num_delta = this->ode_param.num_delta;
+    beta1 = this->ode_param.beta1;
+    beta2 = this->ode_param.beta2;
+    beta3 = this->ode_param.beta3;
+
+    global_size = dealii::Utilities::MPI::sum(storage_register_1.local_size(), this->mpi_communicator);
+    this->pcout << "Allocating ODE system..." << std::flush;
+    this->solution_update.reinit(this->dg->right_hand_side);
+    if(this->all_parameters->use_inverse_mass_on_the_fly == false) {
+        this->pcout << " evaluating inverse mass matrix..." << std::flush;
```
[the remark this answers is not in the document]
jocelynpoon: I think the code still works if use_inverse_mass_on_the_fly == false because I added an abort statement and tried to run the 1D linear advection case and it aborts, however it usually runs

src/flow_solver/flow_solver.cpp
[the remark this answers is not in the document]
jocelynpoon: Ok thanks this is a good idea! I implemented this

src/flow_solver/flow_solver.cpp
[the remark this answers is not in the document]
jocelynpoon: I added this line back in case other cases are impacted.

jocelynpoon:
Hi Carolyn and Shruthi, thanks for going through my code and leaving comments. I went through all your comments and made changes.  Let me know if you see anything else that should be changed, thanks!

src/flow_solver/flow_solver.cpp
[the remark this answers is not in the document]
cpethrick: Sounds good!

src/ode_solver/low_storage_runge_kutta_ode_solver.cpp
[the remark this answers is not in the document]
cpethrick: Thanks! (Note for documentation purposes: we agreed to keep the next few lines commented as they'll be needed for an RRK implementation!)

src/testing/time_refinement_study.cpp
```diff
@@ -116,6 +116,18 @@ int TimeRefinementStudy<dim, nstate>::run_test() const
...
         const int n_timesteps= flow_solver->ode_solver->current_iteration;
         pcout << " at dt = " << dt << std::endl;
+
+
+        if (this->all_parameters->ode_solver_param.runge_kutta_method == PHiLiP::Parameters::ODESolverParam::RK3_2_5F_3SStarPlus 
+            && this->all_parameters->ode_solver_param.atol == 1e-4 && this->all_parameters->ode_solver_param.rtol == 1e-4 
+            && this->all_parameters->time_refinement_study_param.number_of_times_to_solve == 1){
+            double L2_error_expected = 2.14808703658e-5; 
+            pcout << " Expected L2 error is: " << L2_error_expected << std::endl;
+            if (L2_error > L2_error_expected + 1e-16 || L2_error < L2_error_expected - 1e-16){
```
cpethrick, line 126: I would recommend using a slightly larger tolerance (1E-15 maybe) - the tighter tolerance might result in the test failing just due to very small implementation differences which cause rounding to act differently. This is a great test to ensure that the LSRK is working well though!

cpethrick (approved):
Looks good! Thanks for adding that test. I've gone through and checked/resolved my comments.

ssaishruthi98:
I've gone through and checked all my comments and resolved as well! Great work, thanks!

ssaishruthi98 (approved):
I've gone through and resolved all my comments! Great work, thanks! 😄

PR #262, opened 2024-07-02
Update to Narval Compile file for PHiLiP

biondicc:
Small change to compile file to exclude the folders build_release and build_debug rather than all files and folders beginning with build

(without this change the test file build_NNLS_problem.cpp will be excluded and PHiLiP will not compile)


PranshulThakur (approved):
Thanks for the fix. LGTM!

cpethrick (approved):
Looks good! Thanks for the quick fix :)

PR #275, opened 2025-04-24
Documentation improvements for running tests with labels

cpethrick:
A few small changes to make it easier for users to find information about using labels with `ctest`.

### DIFF
--- a/doc/install_ubuntu2004_VM.sh
+++ b/doc/install_ubuntu2004_VM.sh
@@ -183,5 +183,5 @@ mkdir -p Codes
         # 	however, it does not work well on Ubuntu with OpenMPI. Works well with Fedora
         cmake ../ -DCMAKE_BUILD_TYPE=Release -DMPIMAX=4 -DUSE_LD_GOLD=OFF ;\
         make -j4 ;\
-        ctest ;\
+        ctest -LE EXTRA-LONG -LE EXPECTED_FAILURE ;\
 )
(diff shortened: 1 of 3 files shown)


doc/install_ubuntu2004_VM.sh
```diff
@@ -183,5 +183,5 @@ mkdir -p Codes
         # 	however, it does not work well on Ubuntu with OpenMPI. Works well with Fedora
         cmake ../ -DCMAKE_BUILD_TYPE=Release -DMPIMAX=4 -DUSE_LD_GOLD=OFF ;\
         make -j4 ;\
-        ctest ;\
+        ctest -LE EXTRA-LONG -LE EXPECTED_FAILURE ;\
```
cpethrick, line 186: I'm not completely confident we should exclude `EXPECTED_FAILURE` by default. Opinions?

tofstie: Good catch, forgot this file in the original PR

tofstie:
I approve of this change 👍

tofstie (approved):
👍LGTM

doc/install_ubuntu2004_VM.sh
ssaishruthi98: I don't think we should exclude `EXPECTED_FAILURE` by default since we want to ensure that those tests are also behaving as expected.

cpethrick: SG. I will only exclude EXTRA_LONG.

(As an aside, we should work to reduce the number of broken tests...)

PR #276, opened 2025-04-29
Clean-up of RK tableau section

cpethrick:
This PR builds from #267 and unifies the RK tableau classes to have a common base.

There are also some changes to RRK, notably changing from an ODE solver type (`rrk_explicit_solver`) to a bool (`use_relaxation_runge_kutta`). Backwards compatibility has been verified. This change is intended to allow RRK to be used for a more general set of ODE solvers, which may be added to the code in the future.

### DIFF
--- a/src/parameters/parameters_ode_solver.cpp
+++ b/src/parameters/parameters_ode_solver.cpp
@@ -1,4 +1,7 @@
 #include "parameters/parameters_ode_solver.h"
+#include <deal.II/base/mpi.h>
+#include <deal.II/base/utilities.h>
+#include <deal.II/base/conditional_ostream.h>
 
 namespace PHiLiP {
 namespace Parameters {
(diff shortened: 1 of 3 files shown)


src/parameters/parameters_ode_solver.cpp
```diff
@@ -289,6 +295,11 @@ void ODESolverParam::parse_parameters (dealii::ParameterHandler &prm)
             else if (output_string_rrk == "quiet")   rrk_root_solver_output = quiet;
 
             relaxation_runge_kutta_root_tolerance = prm.get_double("relaxation_runge_kutta_root_tolerance");
+            use_relaxation_runge_kutta = prm.get_bool("use_relaxation_runge_kutta");
+            if (use_relaxation_runge_kutta == false && ode_solver_type == rrk_explicit_solver) {
+                // For backwards compatibility
+                use_relaxation_runge_kutta = true;
```
cpethrick, line 301: I could likely remove the `rrk_explicit_solver` before merging this in.

cpethrick: Having thought on it more, I have decided it's best to use `use_relaxation_runge_kutta` primarily, but keep backwards compatibility with the `rrk_explicit_solver` version.

cpethrick: Sure, not a bad idea! It should be completely safe to select the depreciated rrk_explicit_solver, but just in case I have added a warning.

tests/integration_tests_control_files/low_storage_runge_kutta_check/taylor_green_vortex_error_adaptive.prm
```diff
@@ -0,0 +1,69 @@
+# Listing of Parameters
+# ---------------------
+# Number of dimensions
+
+set dimension = 3
+set run_type = flow_simulation #integration_test
```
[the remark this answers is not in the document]
cpethrick: Yes, it will always pass. The test doesn't really verify any functionality other than LSRK not throwing errors for 3D.

PR #277, opened 2025-05-09
Assemble residual on only a  subset of element

cpethrick:
This PR modifies DG such that the residual can be assembled on only a subset of elements.

The vector `list_of_cell_group_IDs` designates a `cell_group_ID` to each cell in the domain. `cell_group_ID` is passed to the `assemble_residual` function, and `assemble_cell_residual` is only called in cells with that ID. By default, the list is initialized to a vector of zeroes, such that the residual is assembled everywhere.

### DIFF
--- a/src/dg/dg_base.cpp
+++ b/src/dg/dg_base.cpp
@@ -1359,6 +1359,24 @@ bool DGBase<dim,real,MeshType>::do_assemble_in_this_cell(const unsigned int cell
     }
 }
 
+
+template <int dim, typename real, typename MeshType>
+void DGBase<dim,real,MeshType>::set_list_of_cell_group_IDs(const dealii::LinearAlgebra::distributed::Vector<int> locations, const int group_ID) {
+
+    // check sizes - not sure if necessary.
+    // bool is_compatible = locations.partitioners_are_compatible(this->list_of_cell_group_IDs);
+
+    // Using only deal.ii vector operations herein to take advantage of their optimizations
+    // Set the cell_group_ID at the given location to zero without changing existing values
+    dealii::LinearAlgebra::distributed::Vector<int> locations_copy((locations));
+    locations_copy.add(-1);
+    locations_copy*=-1;
+    this->list_of_cell_group_IDs.scale(locations_copy);
+    // Set the cell_group_ID at the given location to the group_ID
+    this->list_of_cell_group_IDs.add(group_ID,locations);
+
+}
+
 template <int dim, typename real, typename MeshType>
 double DGBase<dim,real,MeshType>::get_residual_linfnorm () const
 {
--- a/src/dg/dg_base.cpp
+++ b/src/dg/dg_base.cpp
@@ -1088,7 +1088,7 @@ void DGBase<dim,real,MeshType>::reinit_operators_for_cell_residual_loop(
 }
 
 template <int dim, typename real, typename MeshType>
-void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, const bool compute_dRdX, const bool compute_d2R, const double CFL_mass)
+void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, const bool compute_dRdX, const bool compute_d2R, const double CFL_mass, const int cell_group_ID)
 {
     dealii::deal_II_exceptions::disable_abort_on_exception(); // Allows us to catch negative Jacobians.
     Assert( !(compute_dRdW && compute_dRdX)
@@ -1244,7 +1244,7 @@ void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, cons
         if(all_parameters->pde_type == Parameters::AllParameters::PartialDifferentialEquation::physics_model) update_model_variables();
 
         // assembles and solves for auxiliary variable if necessary.
-        assemble_auxiliary_residual();
+        assemble_auxiliary_residual(cell_group_ID);
 
         dealii::Timer timer;
         if(all_parameters->store_residual_cpu_time){
@@ -1253,7 +1253,7 @@ void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, cons
 
         auto metric_cell = high_order_grid->dof_handler_grid.begin_active();
         for (auto soln_cell = dof_handler.begin_active(); soln_cell != dof_handler.end(); ++soln_cell, ++metric_cell) {
-            if (!soln_cell->is_locally_owned()) continue;
+            if (!soln_cell->is_locally_owned() && !this->do_assemble_in_this_cell(soln_cell->active_fe_index(), cell_group_ID)) continue;
 
             // Add right-hand side contributions this cell can compute
             assemble_cell_residual (
@@ -1347,6 +1347,17 @@ void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, cons
     //system_matrix.print(std::cout);
 
 } // end of assemble_system_explicit ()
+
+
+template <int dim, typename real, typename MeshType>
+bool DGBase<dim,real,MeshType>::do_assemble_in_this_cell(const unsigned int cell_index, const int cell_group_ID) const {
+    if (this->list_of_cell_group_IDs[cell_index] == cell_group_ID){
+        // do assemble residual
+        return true;
+    }else {
+        return false;
+    }
+}
 
 template <int dim, typename real, typename MeshType>
 double DGBase<dim,real,MeshType>::get_residual_linfnorm () const
@@ -1895,6 +1906,8 @@ void DGBase<dim,real,MeshType>::allocate_system (
     
     max_dt_cell.reinit(triangulation->n_active_cells());
     cell_volume.reinit(triangulation->n_active_cells());
+
+    list_of_cell_group_IDs.reinit(triangulation->n_active_cells());
 
     // allocates model variables only if there is a model
     if(all_parameters->pde_type == Parameters::AllParameters::PartialDifferentialEquation::physics_model) allocate_model_variables();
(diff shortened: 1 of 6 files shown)


PranshulThakur (commented):
Great PR on assembling residual in a few elements. I didn't find any potential issues and I have just added a few minor comments on efficiency. I think we could use dealii's indexset to reduce the number of if statements and will probably add a more specific comment on it later.

src/dg/dg_base.cpp
```diff
@@ -1896,6 +1940,9 @@ void DGBase<dim,real,MeshType>::allocate_system (
     max_dt_cell.reinit(triangulation->n_active_cells());
     cell_volume.reinit(triangulation->n_active_cells());
 
+    // Unsure whether this should be n_active_cells() or n_global_active_cells()
+    list_of_cell_group_IDs.reinit(triangulation->n_global_active_cells());
```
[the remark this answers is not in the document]
cpethrick: Made that change.

cpethrick:
Currently, all expected `ctest -L QUICK` tests pass. If anyone has comments or concerns about the implementation within `dg`, let me know.

Before opening for review, the following tests should be added to verify behaviour:
* Modify the `assemble_residual_on_subset.cpp` test to have a pass/fail condition: Is the residual indeed assembled only on the subset?
* Add another test that verifies temporal OOA when two partitions are assembled independently. Likely just use forward Euler. May need to verify ghost cell behaviour.

cpethrick:
I ran some timing tests on Narval comparing master and this PR - I wanted to make sure that the extra "if" statement in the cell loop is not causing a significant slowdown. I ran each pair of results on the same Narval node, reserving all cores in the node. While the timings are not as consistent as I may like, there does not seem to be a significant change to the run time.

| |master|PR 277|
|--|----|------|
 -|13354.56398 | 13372.35565
-|13172.13729 | 13311.88947
-|12777.69496 | 13245.94267
-|10112.95472 | 9363.910582
|  |  
Average |12354.33774 | 12323.52459

I'll also run ctest on both to have a better idea of performance on a wider range of tests.
[2D_kelvin_helmoltz_instability.prm.txt](https://github.com/user-attachments/files/20576568/2D_kelvin_helmoltz_instability.prm.txt)
[log-44497360.out.txt](https://github.com/user-attachments/files/20576569/log-44497360.out.txt)

cpethrick:
Opening this for review. I've tidied the comments and added tests. I'm planning to run `ctest` once more, but would like to wait for the narval queue priority to improve a little first.

tofstie (commented):
Thanks for the PR, Here are some preliminary comments while I wait for the tests to be updated

src/dg/dg_base.cpp
```diff
@@ -1253,9 +1253,13 @@ void DGBase<dim,real,MeshType>::assemble_residual (const bool compute_dRdW, cons
 
         auto metric_cell = high_order_grid->dof_handler_grid.begin_active();
         for (auto soln_cell = dof_handler.begin_active(); soln_cell != dof_handler.end(); ++soln_cell, ++metric_cell) {
-            if (!soln_cell->is_locally_owned()) continue;
 
-            // Add right-hand side contributions this cell can compute
+            int mpi_rank;
```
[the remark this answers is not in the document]
cpethrick: thanks for pointing that out! It has been removed in subsequent restructuring.

src/dg/dg_base.cpp
```diff
@@ -551,51 +557,56 @@ void DGBase<dim,real,MeshType>::assemble_cell_residual (
         auto current_face = current_cell->face(iface);
 
         // CASE 1: FACE AT BOUNDARY
```
cpethrick, line 559: The logic for face residual assembly has become quite complicated. I mostly followed the existing structure, but I admit that I've contributed to some messy code!

I'm happy to take functionality suggestions, but I may just open an issue to refactor this section of code. It might be a good task to introduce a new student to the code.

cpethrick:
I have run `ctest -L QUICK` and everything passes except for expected fails. Therefore, I'll open for review and run the full ctest after making any requested changes.

Failing tests from `ctest -L QUICK`: 
```
	 37 - 3D_GMSH_READER_SD7003_04_CELL_SPANWISE (Failed)
	 38 - 3D_GMSH_READER_SD7003_12_CELL_SPANWISE (Failed)
	169 - 1D_ADVECTION_IMPLICIT_MANUFACTURED_SOLUTION_STRONG (Failed)
	170 - 2D_ADVECTION_IMPLICIT_MANUFACTURED_SOLUTION_STRONG (Failed)
	171 - MPI_3D_ADVECTION_IMPLICIT_MANUFACTURED_SOLUTION_STRONG_MEDIUM (Failed)
	249 - ANISOTROPIC_MESH_ADAPTATION_SSHOCK (Failed)
```


(let me know if I should post detailed logs for any of these.)

tofstie:
> I reran some timings on Narval. On a full node and running the same job, I see the following:
> 	master 	PR 277
> - 	15375.957694242 	16179.047770309
> - 	15771.716816927 	39728.893754792
> - 	15925.425937478 	16437.763135127
>   	  	
> Average 	15691 	16308 = 3.9% increase
> 
> Not sure why the second run for this PR was so much longer. I'm assuming that is a problem with the compute node.

Is this just the time it takes to assemble residual or is this the wall clock?

tofstie (approved):
LGTM to me, a small change to a const and a comment about the change in speed

cpethrick:
> > I reran some timings on Narval. On a full node and running the same job, I see the following:
> > master 	PR 277
> > 
> > * ```
> >   15375.957694242 	16179.047770309
> >   ```
> > * ```
> >   15771.716816927 	39728.893754792
> >   ```
> > * ```
> >   15925.425937478 	16437.763135127
> >   ```
> > 
> > Average 	15691 	16308 = 3.9% increase
> > Not sure why the second run for this PR was so much longer. I'm assuming that is a problem with the compute node.
> 
> Is this just the time it takes to assemble residual or is this the wall clock?

This is the wall clock. So that includes residual assembly, overintegrated quantities, etc.

src/dg/weak_dg.cpp
```diff
@@ -3626,6 +3626,33 @@ void DGWeak<dim,nstate,real,MeshType>::assemble_face_term_derivatives(
...
+    dealii::hp::FEValues<dim,dim>                          &/*fe_values_collection_volume*/,
+    dealii::hp::FEValues<dim,dim>                          &/*fe_values_collection_volume_lagrange*/,
+    const dealii::FESystem<dim,dim>                        &/*current_fe_ref*/)
+{
+
+    // Currently for weak DG, do nothing.
+    // This function is needed in strong DG when the current cell is not in an active group
+    // but the neighbor is in an active group. 
+    // Partitioned groups are not currently implemented for weak DG, therefore 
+    // this function does not do anything.
```
PranshulThakur, line 3654: Is this function of WeakDG ever called? if not, maybe we could add an abort statement here.

src/dg/weak_dg.hpp
```diff
@@ -29,6 +29,30 @@ class DGWeak : public DGBaseState<dim, nstate, real, MeshType>
...
+        OPERATOR::basis_functions<dim,2*dim,real>              &/*flux_basis*/,
+        OPERATOR::local_basis_stiffness<dim,2*dim,real>        &/*flux_basis_stiffness*/,
+        OPERATOR::vol_projection_operator<dim,2*dim,real>      &/*soln_basis_projection_oper_int*/,
+        OPERATOR::vol_projection_operator<dim,2*dim,real>      &/*soln_basis_projection_oper_ext*/,
+        OPERATOR::metric_operators<real,dim,2*dim>             &/*metric_oper*/,
+        OPERATOR::mapping_shape_functions<dim,2*dim,real>      &/*mapping_basis*/,
+        std::array<std::vector<real>,dim>                      &/*mapping_support_points*/,
+        dealii::hp::FEValues<dim,dim>                          &/*fe_values_collection_volume*/,
+        dealii::hp::FEValues<dim,dim>                          &/*fe_values_collection_volume_lagrange*/,
+        const dealii::FESystem<dim,dim>                        &/*current_fe_ref*/);
```
PranshulThakur, line 55: Editorial comment: consider using `override`.

src/testing/assemble_residual_on_subset_temporal_OOA.cpp
```diff
@@ -0,0 +1,298 @@
...
+        const double expected_order = 2; // Hard-coded to order = 2
+        const double order_tolerance = 0.1;
+        if (refinement > 0) {
+            L2_error_conv_rate = -log(L2_error_old/L2_error)/log(refine_ratio);
+            pcout << "L2 order at " << refinement << " is " << L2_error_conv_rate << std::endl;
+            if (abs(L2_error_conv_rate - expected_order) > order_tolerance){
+                testfail = 1;
+                pcout << "Expected convergence order was not reached at refinement " << refinement <<std::endl;
+            }
+
```
PranshulThakur, line 274: Not sure how robust an OOA test is to small changes in the residual, such as skipping integration over a face. I think an OOA test should be good enough, but if time permits, one can also consider assembling the entire DG residual, then assembling DG residual for a group ID and then comparing the two vectors to see if the error in residual is close to machine precision.

src/dg/dg_base.cpp
```diff
@@ -612,6 +623,9 @@ void DGBase<dim,real,MeshType>::assemble_cell_residual (
                 const real penalty = 0.5 * (penalty1 + penalty2);
 
                 const dealii::types::global_dof_index neighbor_cell_index = neighbor_cell->active_cell_index();
+            
+                const bool neighbor_cell_is_in_active_cell_group = cell_is_in_active_cell_group(neighbor_cell_index, active_cell_group_ID);
```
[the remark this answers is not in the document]
cpethrick: Yes, so long as the user is careful in assigning the locations (see most recent commit). This is true because the RHS is identical to an unpartitioned assembly on various numbers of cores. I added a comment in the header file to direct a user to example code.

src/testing/assemble_residual_on_subset_temporal_OOA.cpp
cpethrick: Thanks for the suggestion. I updated the unit tests to also assemble a residual without partitioning and compare the values. After some changes to account for ghost cells better, the updated test now passes on serial or MPI.

PR #279, opened 2025-06-20
Paired Explicit RK schemes

jocelynpoon:


DanielDoehring:
I like to see it :heart: :smile:

PR #293, opened 2025-10-28
Comment VTU output on ctest

cpethrick:
A small optimization to ctest to comment the VTU output on some tests. These are not required for verifying the code (e.g., whether ctest passes), and users who are running cases based on the modified .prm files can just uncomment the lines to check output on paraview.

Prior to these changes, about 4 GB of vtu files were generated when running ctest.


DominicRoy2 (approved):
Thanks for doing this!

ssaishruthi98 (approved):
LGTM! Thanks for the making the changes! :smile:

PR #296, opened 2025-11-02
Create Real Gas Constants Object

ssaishruthi98:
Copied over Julien's Masters' code to be used in PHiLiP. This object contains useful functions for multispecies implementation such as GetNASACAP_TemperatureIndex and NASACAP_GetHSCp which use the NASA coefficients to determine real gas properties. This object is standalone and doesn't touch any other part of the code. 

It will be used in the future for implementing a RealGas physics class.

Note: Requires NameOfSpecies.txt and NameOfSpecies.kinetics files which will be included in the tests directory in the future.


ssaishruthi98:
Added as a part of PR #327.

PR #297, opened 2025-11-03
Introduce Number of Species as a compile time const

ssaishruthi98:
To enable the implementation of Real Gas physics in PHiLiP, the number of species used in a simulation is required at compile time to change the maximum value of nstate as required. 

This PR implements nspecies and PHILIP_SPECIES in most classes and CMakeLists. Currently, nspecies and PHILIP_SPECIES are hardcoded to 1, so this should not affect other areas of the code.

Since I was updating almost all templates, I also went the extra step and changed the instantiations in several files to use Boost MPL to reduce the amount of lines in some files (namely, convective_flux.cpp and the physics files). If there is any instantiations I missed that might benefit from Boost MPL, please let me know and I will update that in this PR as well!


cpethrick (commented):
LGTM.

Thanks for making this change self-contained, and for the use of boost! Would you be able to add a comment outlining the template changes everyone else will have to make in their parts of the code when merging this new PR?

I'm happy to approve once other PRs are through, assuming it builds in both release and debug.

PR #331, opened 2026-04-24
Real Gas Entropy Variables

ssaishruthi98:
This PR allows users to run Real Gas simulations using Strong DG. The required functions for Strong DG to run are:

- `compute_entropy_variables`
- `compute_conservative_variables_from_entropy_variables`

Both of which are implemented in this PR. To compute the entropy variables, the following functions were also added:

- `compute_species_entropy`
- `compute_species_gibbs_energy`

Strong DG for Real Gas simulations is verified using the 1D Multi-species Vortex Advection case with 2 species which produces the same orders as the Weak DG run: 
[weak-strong-dg-comparison-ms-va-2s-lt.txt](https://github.com/user-attachments/files/27060445/weak-strong-dg-comparison-ms-va-2s-lt.txt)

The Multi-species Sod Shock Tube test case using Strong DG has also been included in the ctest suite.

Finally, 2 new unit tests have been added! One unit test is implemented to verify the conversion from entropy variables to conservative variables and back for Real Gas. The second unit test is identical to the first but is implemented for Euler. The new unit tests in ctest are:

- `{dim}D_real_gas_convert_entropy_var_conservative`
- `{dim}D_euler_convert_entropy_var_conservative`

### DIFF
--- a/src/dg/dg_factory.cpp
+++ b/src/dg/dg_factory.cpp
@@ -109,6 +109,7 @@ DGFactory<dim,nspecies,real,MeshType>
         }
         else if (pde_type == PDE_enum::real_gas) {
         // nspecies > 1
+        // nspecies > 1
             return std::make_shared< DGStrong<dim,nspecies,dim+nspecies+1,real,MeshType> >(parameters_input, degree, max_degree_input, grid_degree_input, triangulation_input);
         } 
     }
--- a/src/dg/dg_factory.cpp
+++ b/src/dg/dg_factory.cpp
@@ -57,6 +57,7 @@ DGFactory<dim,nspecies,real,MeshType>
 #endif
         }
         else if (pde_type == PDE_enum::real_gas) {
+        // nspecies > 1
             return std::make_shared< DGWeak<dim,nspecies,dim+nspecies+1,real,MeshType> >(parameters_input, degree, max_degree_input, grid_degree_input, triangulation_input);
         } 
     } else {
--- a/src/dg/dg_factory.cpp
+++ b/src/dg/dg_factory.cpp
@@ -106,6 +106,9 @@ DGFactory<dim,nspecies,real,MeshType>
             }
 #endif
         }
+        else if (pde_type == PDE_enum::real_gas) {
+            return std::make_shared< DGStrong<dim,nspecies,dim+nspecies+1,real,MeshType> >(parameters_input, degree, max_degree_input, grid_degree_input, triangulation_input);
+        } 
     }
     std::cout << "Can't create DGBase in create_discontinuous_galerkin(). Invalid PDE type: " << pde_type << std::endl;
     return nullptr;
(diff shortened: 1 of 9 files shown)


src/dg/dg_factory.cpp
```diff
@@ -106,6 +106,9 @@ ::create_discontinuous_galerkin(
             }
 #endif
         }
+        else if (pde_type == PDE_enum::real_gas) {
```
cpethrick, line 110:
```suggestion
        else if (pde_type == PDE_enum::real_gas) {
        // nspecies > 1
```

To make it a little easier to read the logic

src/physics/real_gas.h
```diff
@@ -215,6 +215,11 @@ class RealGas : public PhysicsBase <dim, nspecies, nstate, real>
     // Algorithm 14 (f_M14): Compute species specific internal energy from temperature
     std::array<real,nspecies> compute_species_specific_internal_energy ( const real temperature ) const;
 
+    // Compute species entropy from temperature
```
cpethrick: Could you add a more descriptive comment here? You refer to NASA CAPS in some comments, maybe mention that here

cpethrick (approved):
Looks good. I added two minor comments. 

Should be good to add without a fulll ctest since your changes are specific to real gas. Just the multispecies ctest would be great here.

src/physics/real_gas.h
ssaishruthi98: I added a comment to the Cp, enthalpy and entropy functions explaining that it uses the NASA9 coefficient polynomial parametrization and gave a reference (McBride et al., 2002)! :smile: 

I also added a comment in DG_Factory for real_gas in WeakDG to keep it consistent with the comment added in StrongDG.

ssaishruthi98: Also clarified the meaning of the algorithm and modified by me comments in the RealGas header file! :smile:

src/physics/real_gas.cpp
```diff
@@ -643,6 +623,184 @@ ::compute_species_specific_internal_energy( const real temperature ) const
...
+    return species_entropy;
+}
+
+template <int dim, int nspecies, int nstate, typename real>
+std::array<real,nspecies> RealGas<dim, nspecies, nstate, real>
+::compute_species_gibbs_energy (
+    const std::array<real,nstate> &conservative_soln) const
+{
+    const real temperature = compute_temperature(conservative_soln);
+    const std::array<real,nspecies> species_densities = compute_species_densities(conservative_soln);
```
ohhhershey, line 693: The functions compute_species_gibbs_energy, compute_entropy_variables and compute_conservative_variables_from_entropy_variables are very self explanatory, but having one line comments summarizing them will nice for new readers.

ohhhershey (approved):

src/physics/real_gas.cpp
ssaishruthi98: Thank you, this is a good catch! Writing these comments made me realize that the functions are structured in a way that would only make sense to me and not other users, so I've actually reorganized my functions so that they make more sense and I've included one line comments summarizing them! :smile:

ssaishruthi98:
Here is the multi-species ctest result - 
[LastTestsFailed.log](https://github.com/user-attachments/files/27067908/LastTestsFailed.log)

All passed except two which are expected failures.

PR #333, opened 2026-04-29
Changed file download scripts to point to new GDrive

ssaishruthi98:
A new Google Drive folder has been created for everyone to upload their mesh and initialization files to. This ensures that we have one central location with all the necessary files instead of having the files split across individual user's GDrives.

I've changed the local file download scripts to point to the new GDrive location.

Additionally, I've modified the scripts so that it downloads the entire directory with one command instead of having a separate gdown statement for each file.

I changed the CMake error messages so that they contain the link to the new GDrive as well.


get_flow_initialization_files_local.sh
```diff
@@ -1,15 +1,4 @@
...
-gdown 1oJhZeopd66IbBHepbqlJZWEMNZKECage # setup_philip-00003.dat
-mv setup_philip-00000.dat ${TARGET_DIR}
-mv setup_philip-00001.dat ${TARGET_DIR}
-mv setup_philip-00002.dat ${TARGET_DIR}
-mv setup_philip-00003.dat ${TARGET_DIR}
\ No newline at end of file
+# Download initialization files for decaying homogeneous isotropic turbulence from GDrive
+TARGET_DIR=tests/integration_tests_control_files/decaying_homogeneous_isotropic_turbulence/setup_files/
+gdown --folder 16yJ0mPB8EBLwexKJfNiEsPP40uzx6CHN -O ${TARGET_DIR}
+gdown --folder 1HS7oXwXvFYRCF8CAMV3T3jdR6PBekL9n -O ${TARGET_DIR}
```
PranshulThakur: Should these files be moved to 4proc and 1proc folders?

get_gmsh_mesh_files_local.sh
```diff
@@ -1,29 +1,6 @@
...
-gdown 1RqDeO26q1LRH7s34EEkBiesHJ8yxpswO # 3d_gaussian_bump.msh
-gdown 1MqYCsClOlcm1fVRT0JOSBBzZ1YtB0vZE # 3d_cube_periodic.msh
-gdown 13cjiEOov0tJZGA1DNmq5B36Qycdyakl0 # SD7003_1_cell_spanwise.msh
-gdown 1nLO4RCUDcM6Mt_jBkQdEc-6KlhT-WpxI # SD7003_4_cell_spanwise.msh
-gdown 1fsU531h3aEMy34O16yf8WODFxBIkY_Wm # SD7003_12_cell_spanwise.msh
-# move files to appropriate directory
-mv *.msh ${TARGET_DIR}
+# Download necessary test files from GDrive
+# Mesh files for integration tests
+gdown --folder 1QOYxwyo4MMZasTiRnayHCnZ4E8oMR0gL -O tests/
```
PranshulThakur: Should be -O tests/meshes/ ?

get_gmsh_mesh_files_local.sh
```diff
@@ -1,29 +1,6 @@
...
-gdown 1RqDeO26q1LRH7s34EEkBiesHJ8yxpswO # 3d_gaussian_bump.msh
-gdown 1MqYCsClOlcm1fVRT0JOSBBzZ1YtB0vZE # 3d_cube_periodic.msh
-gdown 13cjiEOov0tJZGA1DNmq5B36Qycdyakl0 # SD7003_1_cell_spanwise.msh
-gdown 1nLO4RCUDcM6Mt_jBkQdEc-6KlhT-WpxI # SD7003_4_cell_spanwise.msh
-gdown 1fsU531h3aEMy34O16yf8WODFxBIkY_Wm # SD7003_12_cell_spanwise.msh
-gdown 11loBYH22aArwIF78tJzN-YjtzZhri-7p # channel_structured.msh
-# move files to appropriate directory
-mv *.msh ${TARGET_DIR}
\ No newline at end of file
+gdown --folder 1Z89Qq940bP9WwBr1Y5YVhZSAErOM0Rg6 -O tests/unit_tests/grid/
```
PranshulThakur: Move to tests/unit_tests/grid/gmsh_reader/?

PranshulThakur (commented):
Thanks for adding these files to a common drive. I just have minor comments, and other than that, everything looks good.

get_flow_initialization_files_local.sh
ssaishruthi98: The folder on GDrive is called 1proc and 4proc respectively. The `--folder` modifier for the gdown command indicates it should download the entire folder, so that command is downloading `1proc` to `tests/integration_tests_control_files/decaying_homogeneous_isotropic_turbulence/setup_files/` (same applies to 4proc).

get_gmsh_mesh_files_local.sh
ssaishruthi98: The folder on GDrive is called meshes. The --folder modifier for the gdown command indicates it should download the entire folder, so that command is downloading `meshes` to `tests/`.

get_gmsh_mesh_files_local.sh
ssaishruthi98: The folder on GDrive is called gmsh_reader. The --folder modifier for the gdown command indicates it should download the entire folder, so that command is downloading `gmsh_reader` to `tests/unit_tests/grid/`

PranshulThakur: Ok, I don't see any major issues, other than if someone has their specific mesh file in those folders, it would get deleted. We usually pull mesh files just once, so it should be good.

PranshulThakur (approved):
LGTM!

get_gmsh_mesh_files_local.sh
ssaishruthi98: The expected behaviour is that it wouldn't delete any existing files. For example, tests/meshes has a CMakeLists.txt but the CMakeLists file is still there after the script is run. 

When I tested it on my computer that is what happened but I would appreciate it if you could double check this on your end! 😄

PranshulThakur: Interesting. I tested it and it does not delete existing files. Looks like gdown has implemented checks for not deleting the folder.

ssaishruthi98: Awesome, good to hear! Thank you 😄

DominicRoy2 (approved):
Thank you for cleaning this up!

ssaishruthi98:
Based on some feedback, I have:

- condensed the two different file download scripts into one script for local PHiLiP purposes.
- created one unified script for both rorqual and narval test file downloads instead of having 4 different scripts. The new script prompts the user to enter the name of the cluster and sets the path_to_files accordingly. It is case insensitive as well.

The local script changes have been tested but since rorqual and narval are down today we cannot test the cluster changes. The files will be moved into new directories on Monday and once testing is complete, the PR can be reviewed.

ssaishruthi98:
While testing the cluster scripts I checked to see if we can use `gdown` on the clusters and it turns out we can! So in light of this discovery, I have deleted the cluster script because we can just use the local script to download mesh files for Narval and Rorqual as well.

The user will have to `pip install gdown` but that's something we have to do locally anyways so this change shouldn't be an issue! 

@PranshulThakur and @DominicRoy2 I think this PR is ready for your re-review 😄

PranshulThakur (approved):

DominicRoy2 (approved):
Thank you for cleaning this up!

PR #335, opened 2026-05-20
Bugfix for domains where a boundary cell has no neighbor

cpethrick:
I encountered a bug while updating one of my branches. Strong DG segfaults if there is not a neighbor cell (eg, if the domain is 1 cell wide perpendicular to a boundary cell). Changed behaviour to return the current cell and a dummy face index when not using a wall model.

I verified that the two wall model tests still work - see attached logfile:
[log_WALL_tests.txt](https://github.com/user-attachments/files/28064623/log_WALL_tests.txt)


ssaishruthi98 (approved):
Thanks for the fix! The changes look good to me 😄

DominicRoy2 (approved):
Thank you for this bug fix! Looks good to me.

PR #336, opened 2026-05-26
Change CMake error messages to refer to new script name

cpethrick:
Follow up from #333 

The script name was changed, but when the user does not have mesh/initialization files, the error message pointed to the old script names.


cpethrick:
I recall that the scripts work on the cluster - @ssaishruthi98 can you confirm in your review?

ohhhershey (approved):

ssaishruthi98 (approved):
Thanks for catching that! The script works on the clusters. The changes look good to me. 😄

Issue #326, opened 2026-03-20
Force user to choose a Reynolds Number when using Navier-Stoke Physics


ssaishruthi98:
Currently there is no code in place to ensure that the user selects a Reynolds Number before running the test case.

The default Re is 10000000. Ideally, this would be modified such that the use of the default value would lead to an error with a message directing the user to set Re.

This would probably require the default value to be set to -1.0 and then a check being included in the parameter processing.

PR #337, opened 2026-06-01
Issue #326

gillianhelliwell-boop:
Solution to issue #326 that involves changing the default Re=0 and adding a check for Navier Stokes in physics_factory.cpp. I checked all Navier-Stokes test cases an Euler & Burgers test case and they all passed.


src/parameters/parameters_navier_stokes.cpp
```diff
@@ -11,8 +11,8 @@ void NavierStokesParam::declare_parameters (dealii::ParameterHandler &prm)
                           dealii::Patterns::Double(1e-15, dealii::Patterns::Double::max_double_value),
                           "Prandlt number. Default value is 0.72. "
                           "NOTE: Must be consitent with temperature_inf.");
-        prm.declare_entry("reynolds_number_inf", "10000000.0",
-                          dealii::Patterns::Double(1e-15, dealii::Patterns::Double::max_double_value),
+        prm.declare_entry("reynolds_number_inf", "0", //gillian changed to 0 to address issue
```
ssaishruthi98: I think the comment can be removed here.

src/physics/physics_factory.cpp
```diff
@@ -124,7 +124,12 @@ ::create_Physics(const Parameters::AllParameters                              *c
                     diffusion_tensor, 
                     manufactured_solution_function);
         } else if (pde_type == PDE_enum::navier_stokes) {
+            if (parameters_input->navier_stokes_param.reynolds_number_inf == 0) {     // 0 is the default Reynold's number
+                     std::cout << "Error! No Reynold's number was provided." << std::endl;
+                     std::abort();
+                }
             if constexpr (nstate==dim+2) {
+
```
ssaishruthi98: I think the newline should be deleted here as it is best practice to not add newlines where it is not necessary.

ssaishruthi98 (commented):
Great work on your first PR Gillian! 😄 I just have two minor comments and I can approve it after. Although given the type of change (change to parameters) it might be best to do a ctest run and post the results as well since that is usually standard for our PRs.

ssaishruthi98 (approved):
The changes look good to me, thanks Gillian! 😄

src/physics/physics_factory.cpp
```diff
@@ -124,6 +124,10 @@ ::create_Physics(const Parameters::AllParameters                              *c
                     diffusion_tensor, 
                     manufactured_solution_function);
         } else if (pde_type == PDE_enum::navier_stokes) {
+            if (parameters_input->navier_stokes_param.reynolds_number_inf == 0) {     // 0 is the default Reynold's number
+                     std::cout << "Error! No Reynold's number was provided." << std::endl;
```
cpethrick: Small correction - no apostrophe in Reynolds! It's named for Osborne Reynolds :)

gillianhelliwell-boop: thanks!

src/flow_solver/flow_solver_cases/periodic_turbulence.cpp
```diff
@@ -47,7 +47,11 @@ PeriodicTurbulence<dim, nspecies, nstate>::PeriodicTurbulence(const PHiLiP::Para
     
     // Navier-Stokes object; create using dynamic_pointer_cast and the create_Physics factory
     using PDE_enum = Parameters::AllParameters::PartialDifferentialEquation;
-    PHiLiP::Parameters::AllParameters parameters_navier_stokes = this->all_param;
+    Parameters::AllParameters all_parameters_write = (this->all_param); //pointer with write permission for Reynolds number declaration
+    if(this->all_param.pde_type == PDE_enum::euler) {
+        all_parameters_write.navier_stokes_param.reynolds_number_inf = 10000000;  //added arbitrary default to prevent test case failure
```
sivanadarajah, line 52: I can't remember why are we setting a default for Euler! Which test fails due to this? Perhaps we need to fix the test.

PR #338, opened 2026-06-09
Skip entropy projections when using collocated nodes

ssaishruthi98:
Currently, StrongDG calls the functions `compute_entropy_variables`, `compute_conservative_variables_from_entropy_variables` and uses projected entropy variables even when using collocated nodes. 

These steps are unnecessary and result in an increased simulation time and increased cost. 

This PR introduces a flag in StrongDG that stores whether the nodes are uncollocated and the flag is used to determine whether entropy variables projections are obtained or not. 

Additionally, while implementing this, I noticed the `use_projected_entropy_variables_for_nsfr_boundary_term` is used to determine whether we pass the solution interpolated to face or projected entropy variables to the `boundary_face_values` function but no test case uses this prm and the users aren't aware of it's existence either. Since we want to pass the projected entropy variables in case of GL nodes and solution interpolated to face in case of GLL nodes, I've removed that prm entirely and replaced it with the new flag I've introduced.


src/dg/strong_dg.cpp
```diff
@@ -2045,43 +2053,45 @@ void DGStrong<dim,nspecies,nstate,real,MeshType>::assemble_boundary_term_strong(
 
     // First, transform the volume conservative solution at volume cubature nodes to entropy variables.
     std::array<std::vector<adtype>,nstate> entropy_var_vol;
```
sivanadarajah, line 2055: Where is this vector used?

ssaishruthi98: This is used to obtain `entropy_var_coeff` on line 2082 which is then used to obtain `projected_entropy_var_vol`.

ssaishruthi98:
I have completed a ctest run for this PR. I have attached the results. All are expected failures except for the TGV cases which pass locally.

The Dipole cases are failing as a result of the changes in this branch but the fix for these cases would be to investigate the viscous wall BC for GL nodes (using entropy projected variables for the boundary condition function). A separate issue has been created for this (#341).
[LastTestsFailed.log](https://github.com/user-attachments/files/29145982/LastTestsFailed.log)

DominicRoy2 (approved):
Thank you for implementing this change!

sivanadarajah (approved):
Excellent work.

PR #339, opened 2026-06-09
Diagonally-Implicit RK with AD exact Jacobian

cpethrick:
Draft - needs cleaning before it's ready to merge.


src/ode_solver/runge_kutta_ode_solver.h
```diff
@@ -37,7 +37,9 @@ class RungeKuttaODESolver: public RungeKuttaBase <dim, nspecies, real, n_rk_stag
     /// Function to adjust time step size
     real adjust_time_step (real dt) override;
 
-protected:
+    std::array<dealii::LinearAlgebra::distributed::Vector<double>,3> soln_stored;
```
cpethrick, line 40: This causes a segfault if n_rk_stages > 3. Should be removed unless it is used anywhere.

PR #340, opened 2026-06-10
Adding far-field boundaries to navier_stokes

DominicRoy2:
Adding existing far-field boundaries to navier_stokes.cpp. This avoids issues with the auxiliary equation for the boundary term, where the solution at the boundary is not initialized. This bug was introduced with the new viscous boundary conditions.


ohhhershey (approved):

ssaishruthi98 (commented):
Everything looks good to me! Thank you for the change 😄 

I do think it would be good to run a full `ctest` (or at least a `ctest -L NAVIER_STOKES` to ensure that existing viscous cases don't fail because of the change.

DominicRoy2:
Results from a Rorqual ctest run are attached. 
[failed-tests-pr340.txt](https://github.com/user-attachments/files/30354851/failed-tests-pr340.txt)

ssaishruthi98 (approved):
Awesome, thank you! 😄

PR #343, opened 2026-07-27
New Test Tag - KNOWN_FAILURE

ssaishruthi98:
When PR #338 is merged into master, the dipole wall collision test cases are expected to fail since we are no longer passing the solution interpolated to the face to the boundary function when using GL nodes. The dipole cases collectively take around 12.5 hours to run on rorqual using 16 cores and only fail at the very end of the test when the final values are verified against the expected values.

To make the PR process more efficient and ensure we don't waste our resources, I have added a tag for tests called `KNOWN_FAILURE`. This tag is added to test cases that have issues open on GitHub. This will allow us to skip the tests we know will not pass.

The command for running ctest locally while excluding `EXTRA-LONG` and `KNOWN_FAILURE` would be `ctest -LE 'EXTRA-LONG|KNOWN_FAILURE'`


ssaishruthi98:
Will change this PR from draft to ready for review once PR #338 is merged.

cpethrick:
I think it's okay to add this label, but with the caveat that ctest runs on a compute cluster should run all tests. Locally, I think it's okay to omit known failing tests and extra-long tests so long as we are regularly testing all tests on the cluster.

Could you please update the documentation to indicate the command `ctest -LE 'EXTRA-LONG|KNOWN_FAILURE'` ?

